PR #11, opened 2021-12-15
endpointDefinition => LazyLoad

### LINKED ISSUE #10: Unit Tests randomly fail
Hi,

I am some issue with this module.   It seems that when calling 
 app.UseFastEndpoints(); 

Tests are randomly failing with Error:
``The server has not been started or no web application was configured.``

Culprit:
[Task.Run(async () => ](https://github.com/dj-nitehawk/FastEndpoints/blob/dcdad6631acd95ac0c7763a47ed2c2fa82952065/Src/Library/Extensions/MainExtensions.cs#L118)
It seems that after 1 minute this static variables (**discoveredEndpointDefinitions**) gets set to null

I know that I am using **Specflow**/**xUnit** and that a the WebApplicaitonFactory is created for every scenario;  I tried disabling the parallelism, but tests run too slow.

Why is the code doing that ?   And could we make it configurable or be able to disable that behavior ?

bobbyangers:
thread safety
FIX #10

### DIFF
--- a/Src/Library/Extensions/MainExtensions.cs
+++ b/Src/Library/Extensions/MainExtensions.cs
@@ -18,19 +18,7 @@ namespace FastEndpoints;
 /// </summary>
 public static class MainExtensions
 {
-    private static Lazy<List<EndpointDefinition>> endPointDefinitions = new(() =>
-    {
-        var sw = new Stopwatch(); sw.Start();
-        var result = GenerateEndpointDefinitions();
-        endpointDiscoveryDurationTicks = sw.ElapsedTicks;
-        sw.Stop();
-
-        if (result is null || result.Length == 0)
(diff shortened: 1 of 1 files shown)


bobbyangers:
Check this technique, here I am using a Lazy<T> for the endpointDefinition.   This construct implement proper mutex to guarantee thread safety.  (which is not when using a static contructor) unless you use a 
`` lock( someMutex ) {
} ``

dj-nitehawk:
great will check it out first thing tomorrow morning. does it now solve
your initial problem?

dj-nitehawk:
released v2.12.0-beta1 to nuget. give it a go and let me know how it performs...
thanks for the contribution 🙏👍

PR #37, opened 2022-01-28
Fix enum parsing

vfrz:
`Type.GetTypeCode` on an enum type doesn't return `TypeCode.Object` but the TypeCode of the enum's "sub type", so `int` by default.


dj-nitehawk:
thank you!

PR #47, opened 2022-02-28
Add doc on how to model bind raw request content

fauresco:
Adding documentation on how to bind to the raw request. I was going to try to implement something myself when I discovered the IPlainTextRequest interface already exists! Awesome! :)

I have just changed the Model-Binding.md file because I have no experience with the docfx. I have tried to run it but it changed a lot of files (not just version number, it changed the actual contents), so I am not sure if I am missing some command line arguments or doing something wrong. Please let me know if you want me to run it, and if so, what is the correct process.


dj-nitehawk:
looks like it missed my radar :-)
appreciate the PR...
👍

PR #51, opened 2022-03-05
add overload sendokasync to accept response body

cjamesrohan:
this allows an overload to SendOkAsync to take a body. similar to return Ok() and return Ok(response) that most .net devs would be familiar with.


dj-nitehawk:
merged. but i still don't see the point of `SendOkAsync(dto)` vs `SendAsync(dto)` as they both do the exact same thing. other than the fact that one call is 2 chars shorter than the other.
no big deal ;-)

PR #97, opened 2022-04-26
Example request bodies in swagger

hoppel:
Very basic implementation of custom example request bodies in swagger, will only work for simple complex models, if they use BindFrom etc it will not work properly.


dj-nitehawk:
i just discovered that we can simply do the following:
```csharp
public class MyRequest
{
    /// <example>example username</example>
    public string UserName { get; set; }

    /// <example>example password</example>
    public string Password { get; set; }
}
```
is this not good enough? do we really need to complicate things and waste memory to hold an example object when the same end result can be achieved easily with the above?

what do you think? should we scrap this change?

hoppel:
Personally I prefer being be able to just create a type and use that instead, especially for larger requests with more complex objects. 
Perhaps both can be documented with the focus on using XML for applications that need to achieve low memory?

hoppel:
Hey @dj-nitehawk looking at this now, what would your preferred way of implementing this? I was thinking about using a custom `DefaultContractResolver` which we can filter out properties based on attributes but I'm not sure if this would collide with `SchemaGenerator.Settings.ActualSerializerSettings` which is being used currently

dj-nitehawk:
most likely, a custom contract resolver would be the way to go. but i think it would have to copy the code from here: https://github.com/RicoSuter/NJsonSchema/blob/master/src/NJsonSchema/Generation/SystemTextJsonUtilities.cs
because it's a private sealed class.

which is being used to configure swagger here: https://github.com/dj-nitehawk/FastEndpoints/blob/main/Src/Swagger/Extensions.cs#L72

what do you think?

hoppel:
Perhaps we could try a PR so we dont have to copy it?
Will copy it for now and play around with it to see if I can get it working

dj-nitehawk:
yeah sounds good!

hoppel:
My first thought was to just do all the filtering in the resolver again but we already have a "cleaned" model for the default request body, we could just select those properties and remove all properties from the example body if they dont exist in the default one?
(speaking about the body thats passed into this method multiple times for cleanup https://github.com/dj-nitehawk/FastEndpoints/blob/main/Src/Swagger/OperationProcessor.cs#L342 )

dj-nitehawk:
that certainly is an alternative. what about storing the property names in a list everytime  `RemovePropFromRequestBodyContent()` and use that list to match/filter the props that shouldn't exist in the end result?
or we take the property values from the user supplied object and set them directly on the `op.RequestBody.Content`. not sure if it's possible though.

dj-nitehawk:
i think i've cracked it in `v4.2.0-beta6`

could you give it a good test run pls?

dj-nitehawk:
fixed a few issues in `v4.2.0-beta7`

hoppel:
Your approach is definitely a lot fancier than the mess I came up with haha. It looks good, the only issue I have im using a custom JsonConverter which is not being respected when using the summary example.

PR #98, opened 2022-04-28
Add test case for post with empty body

fretje:
and a failing one for how nswag client is handling it.

See https://discord.com/channels/933662816458645504/968858009520451595/969261166599749652 for context.


dj-nitehawk:
did a benchmark with a potential solution which doesn't seem to be performant.
see here: https://github.com/dj-nitehawk/FastEndpoints/issues/124#issuecomment-1133938881

PR #103, opened 2022-05-04
Log the entire exception

Toyotomi-clan:
Logging the entire exception enables tools like serilog.exceptions to enrich exceptions.


dj-nitehawk:
pls make this an optional/opt-in feature. let's keep just the stack trace by default and let ppl who want the whole exception logged pass in their intent via method argument.

Toyotomi-clan:
@dj-nitehawk I moved this to [this pull request](https://github.com/dj-nitehawk/FastEndpoints/pull/105)

PR #104, opened 2022-05-04
Fix Route parameter property type not correctly set with swagger

meysamhadeli:
Fix Route parameter property type not correctly set with swagger


dj-nitehawk:
thank you for your contribution 🙏🙏🙏

PR #109, opened 2022-05-07
Resolves #86

### LINKED ISSUE #86: Body is removed from schema if all non-inherited properties are marked as QueryParams
I'm using FastEndpoint 4.0.

In some cases in our code there are base classes used for sharing properties. However if an endpoint has all non-inherited properties marked as QueryParams, the body is removed from the schema even if there are inherited properties remaining.

Example:
```csharp

public class EndpointError : Endpoint<RequestClass, EmptyResponse>
{
    {
        Post("/error");
    }

    public override async Task HandleAsync(RequestClass req, CancellationToken ct)
    {
        await SendOkAsync(cancellation: ct);
    }
}

public class BaseClass
{
    public string InheritedProp { get; set; }
}

public class RequestClass : BaseClass
{
    [QueryParam]
    public string Prop1 { get; set; }
    // [QueryParam] // uncomment this to reproduce the error
    public string Prop2 { get; set; }
}
```

This is OK:
image

But if `Prop2` is also a QueryParam, we suddenly get this:
image

I also checked the swagger generated, just to be sure, and the schema is missing there as well.

emrekara37:
Hi,
If there is no property in the model, not showing in swagger
#86

### DIFF
--- a/Src/Swagger/OperationProcessor.cs
+++ b/Src/Swagger/OperationProcessor.cs
@@ -271,6 +271,7 @@ internal class OperationProcessor : IOperationProcessor
         //remove request body if there are no properties left after above operations
         //otherwise there's gonna be an empty schema added in the swagger doc
         if (op.RequestBody?.Content.SelectMany(c => c.Value.Schema.ActualSchema.ActualProperties).Any() == false &&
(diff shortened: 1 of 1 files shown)


dj-nitehawk:
thanks buddy!
much appreciated 🙏🙏🙏

Issue #78, opened 2022-03-28
Add complex model binding to form-data requests


roel-de-vries:
As stated in [the docs](https://fast-endpoints.com/wiki/Model-Binding.html#complex-model-binding), "complex model binding is only supported from the json body". My use-case is sending files together with other data. I could live with flattening my request models, but this also prevents the posiblity of sending an object array.

Something like sending the following request is not possible using FastEndpoints:
```csharp
public class PersonCreateRequest
{
    public string Name { get; set; }
    public IFormFile Picture { get; set; }
    public List<Car> Cars { get; set; }
}

public class Car 
{
    public string LicensePlate { get; set; }
    public int Mileage { get; set; }
}
```

Whereas this .NET 6 api endpoint
```csharp
[ApiController]
[Route("[controller]")]
public class PersonController : ControllerBase
{
    [HttpPost]
    public IActionResult Get([FromForm] PersonCreateRequest req)
    {
        return Ok();
    }
}

public class PersonCreateRequest
{
    public string Name { get; set; }
    public IFormFile Picture { get; set; }
    public List<Car> Cars { get; set; }
}

public class Car
{
    public string LicensePlate { get; set; }
    public int Mileage { get; set; }
}
```

Does provide support for handling the following form-data request:
![image](https://user-images.githubusercontent.com/2130723/160406438-c4e0aedc-beff-43ec-b79a-72493f763e0d.png)

Having the posibility to use complex model binding also with form-data would a great addition I think.

dj-nitehawk:
it certainly would be a great addition. however i do not currently have the capacity to look in to this for a few weeks. so i'll mark this as `contribution needed` so someone else could possibly look in to sending in a PR.

in the meantime, you could send the cars as json strings and do the following:

![screen-record-1648481893](https://user-images.githubusercontent.com/7043768/160435148-4114cd55-3490-48f7-a37e-edb9973883a8.png)

```csharp
public class CreatePerson : Endpoint<PersonCreateRequest>
{
    public override void Configure()
    {
        Post("create-person");
        AllowAnonymous();
        AllowFileUploads();
    }

    public override Task OnBeforeHandleAsync(PersonCreateRequest req)
    {
        req.Cars = Form["Cars"].Select(c => JsonSerializer.Deserialize<Car>(c)).ToList();
        return Task.CompletedTask;
    }

    public override async Task HandleAsync(PersonCreateRequest req, CancellationToken ct)
    {
[...]

PR #110, opened 2022-05-07
Array or List binding over the FormData #78

emrekara37:
Hi,
#78 
I tried with the data in this issue it working

### DIFF
--- a/Src/Library/Extensions/ReflectionExtensions.cs
+++ b/Src/Library/Extensions/ReflectionExtensions.cs
@@ -1,6 +1,9 @@
-﻿using System.Collections.Concurrent;
+﻿using Microsoft.AspNetCore.Components.Forms;
+using System.Collections;
+using System.Collections.Concurrent;
 using System.Linq.Expressions;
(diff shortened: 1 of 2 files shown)


dj-nitehawk:
so, for it to work, the incoming data should be a string containing a json object array right?
also what if we changed the match condition to `IEnumerable` instead of `IList`? would that cover a larger range of property types?

emrekara37:
> so, for it to work, the incoming data should be a string containing a json object array right? also what if we changed the match condition to `IEnumerable` instead of `IList`? would that cover a larger range of property types?


Yes, For example this issue #78  

```csharp
var cars= Form["Cars"] // {"Mileage":1,"LicensePlate":"A"} , {"Mileage":2,"LicensePlate":"B"} , {"Mileage":3,"LicensePlate":"C"}
```

And i added just  **[  ]** 


Yes i think we can change IEnumerable instead of IList,

@dj-nitehawk

dj-nitehawk:
great! thank you for the contribution 🙏🙏🙏

PR #111, opened 2022-05-07
Adding some unit tests

mehdihadeli:


Tests/IntegrationTests/IntegrationTests.Shared/Fixtures/IntegrationTestFixture.cs
```diff
@@ -1,82 +1,6 @@
...
 
-public class IntegrationTestFixture : IAsyncLifetime
+public class IntegrationTestFixture
 {
-    private readonly CustomWebApplicationFactory<Program> _factory;
-
-    public IntegrationTestFixture()
-    {
-        //hack for being able to run integration tests in azuer/github containers
-        Environment.SetEnvironmentVariable("DOTNET_hostBuilder:reloadConfigOnChange", "false");
```
dj-nitehawk, line 18: @mehdihadeli you will have to set this environment variable before any testhosts/factories are created. without this, the pipelines in github/azure are crapping out. i don't really know exactly why it's happenning. maybe a because a testhost is being created per test execution? i'm not too familiar with xunit, is there a hook/place where this env-var can be set on xunit initialization/startup?

mehdihadeli: I put your mentioned config in the correct place.

PR #118, opened 2022-05-16
Update error response so that error code and the severity can be retu…

philippemercure:
…rn to the client when set in code.

This would allow to have an error code and severtity return to the client, so action could be taken accordingly, without relying on the error message only.

### DIFF
--- a/Src/Library/DTOs/ErrorResponse.cs
+++ b/Src/Library/DTOs/ErrorResponse.cs
@@ -20,7 +20,7 @@ public sealed class ErrorResponse
     /// <summary>
     /// the collection of errors for the current context
     /// </summary>
-    public Dictionary<string, List<string>> Errors { get; set; } = new();
+    public Dictionary<string, List<ErrorResponseObject>> Errors { get; set; } = new();
 
     /// <summary>
     /// instantiate a new error response without any errors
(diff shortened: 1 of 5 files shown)


dj-nitehawk:
sorry mate. i don't think this PR is acceptable as it's going to break every client out there.
and i certainly don't want my default error responses to look like that. i'm sure most others won't either.

however, i will accept a PR with just the following changes:

https://github.com/dj-nitehawk/FastEndpoints/pull/118/files#diff-f1cda060f2e717194a633f1e00b80e275fb6810171ae0c430bba316baa419b15R14-R25

https://github.com/dj-nitehawk/FastEndpoints/pull/118/files#diff-f1cda060f2e717194a633f1e00b80e275fb6810171ae0c430bba316baa419b15R32-R42

that way someone who wants to do this can build their own custom error response as described here: 
https://fast-endpoints.com/wiki/Configuration-Settings.html#customizing-error-responses

philippemercure:
@dj-nitehawk, just to let you know that in Rider, the nuget package 4.2.0-beta10 is consider a downgrade, I think it's because of the name. beta10 is between beta 1 and beta 2. This does not happen In Visual Studio.

dj-nitehawk:
i guess that's a bug report for rider then ;-)

PR #119, opened 2022-05-17
add template pack with tests

### LINKED ISSUE #107: Create a template pack for `dotnet new`
I like the idea of the VS extension for easily creating the recommended files for slices. Since many people these days don't use VS for .NET development, I think it would be helpful to create a template pack that can be used with `dotnet new` to create the same files that the extension is creating, especially for those who are just making the switch from controllers to slices. 

I'm pretty fluent with template packs, so if that sounds good I can start working on it.

lazyboy1:
Created template pack with multiple options (too much? 🙂), based on the VS extension item template. It also includes a tests project for the different optional parameters.
The `TemplatePack` project needs to be published to nuget.org.

Closes #107 

usage:
```
> dotnet new feat --help

FastEndpoints Feature Fileset (C#)
Author: @lazyboy1
Options:
  -t|--attributes  Whether to use attributes for endpoint configuration
                   bool - Optional
                   Default: false

  -p|--mapper      Whether to use a mapper
                   bool - Optional
                   Default: true

  -v|--validator   Whether to use a validator
                   bool - Optional
                   Default: true

  -m|--method      Endpoint HTTP method
                       GET
                       POST
                       PUT
                       DELETE
                       PATCH
                   Default: GET

  -r|--route       Endpoint path
                   string - Optional
                   Default: api/route/here
```


dj-nitehawk:
looks pretty good...
will push to nuget soon...
thanks buddy!

PR #120, opened 2022-05-18
Update AddError method to have an optionnal ErroCode and Severity tha…

philippemercure:
…t can be set.

This would allow to have an error code and severity set and then set with the CustomizeResponseError : https://fast-endpoints.com/wiki/Configuration-Settings.html#customizing-error-responses


philippemercure:
@dj-nitehawk When can I expect this in beta10? Thanks.

Issue #117, opened 2022-05-15
using source generators to cut down on endpoint/type discovery time during startup


dj-nitehawk:
currently fastendpoints uses assembly scanning to discover endpoints/validators/event handlers/summaries during startup.
https://github.com/dj-nitehawk/FastEndpoints/blob/1745ec0254da63d22bfea6d96067d047f0b8943e/Src/Library/Extensions/EndpointData.cs#L54-L70
this obviously takes some milliseconds whenever the app is run, which would not be ideal for serverless/cold start scenarios.

if we can use a source generator to produce a static class with a `System.Type[]` static property at compile time, we may be able to get rid of the type discovery time.

the end result of the source generation should be something like this:
```csharp
namespace FastEndpoints
{
    public static class DiscoveredTypes
    {
        public static readonly System.Type[] AllTypes = new System.Type[]
        {
            typeof(UsersNameSpace.MyEndpoint1),
            typeof(UsersNameSpace.MyEndpoint2),
            typeof (UsersNameSpace.MyEndpoint3)
        };
    }
}
```
if you have experience with source generators, please see if you can implement this.
this will be much appreciated by the folk who uses fastendpoints in serveless environments.

i myself got as far as trying to figure out if a `ClassDeclarationSyntax` implements a given interface such as `IEndpoint` and failed since it seems like it's not possible to get a list of interfaces a class implements inside the `SyntaxReceiver`.

nefarius:
Started working on this in [a new branch](https://github.com/nefarius/FastEndpoints/tree/nefarius/feature/source-generators).

nefarius:
Looks promising 😄 

![image](https://user-images.githubusercontent.com/286631/169362017-fa5062eb-03eb-4f00-8e37-2bd3b2080ca5.png)

dj-nitehawk:
@nefarius forgot to tell you... i had to change the assembly discovery logic in order to cater for #121
pls pull those changes into your fork/branch.

nefarius:
@dj-nitehawk so you need a property returning a list of assemblies instead of types, did I get that right?

dj-nitehawk:
nevermind... looks like you've already pulled the latest code:
https://github.com/nefarius/FastEndpoints/blob/nefarius/feature/source-generators/Src/Library/Extensions/EndpointData.cs

carry on... 😁

nefarius:
[...]

PR #125, opened 2022-05-20
Added source generator package for endpoint discovery

nefarius:
This implements #117 

This PR adds another project/NuGet package with a code analyzer the consumer of the library can additionally use in their project. Upon building, a static utility class with all discovered types gets generated, which in turn will be used by the main library - if available - instead of using reflection. If the package is not used, the current method with reflection is run.


dj-nitehawk:
looks good...
let me give it a good try...
thank you very very much for this!!!

PR #130, opened 2022-05-23
Some source generator fixes

nefarius:
- Fixed a case where types under the global namespace were ignored
- Reworked filtering against interfaces of interest
- Bumped `Microsoft.CodeAnalysis.CSharp` version to `4.2.0`


dj-nitehawk (approved):
looks good. will merge and do some testing.

PR #137, opened 2022-05-24
Allows multiple IvalidationRules per property

### LINKED ISSUE #136: [swagger-fv-integration] multiple validation rules for the same dto property
if two rules are present for the same property like this:

https://github.com/dj-nitehawk/FastEndpoints/blob/f141f7b413b207be176d244814eed4d7b51e2534/Web/%5BFeatures%5D/Admin/Login/Models.cs#L39-L47

the following exception is thrown.

> System.ArgumentException: 'An item with the same key has already been added. Key: UserName'

@wjax since this is a valid way to write validation rules, how should we handle this?

**some options:**
1. find a way to combine the rules into one and add that to the dictionary (might not be feasible)
2. take only the first rule
3. throw an exception and ask the user to combine multiple rules into one themselves
4. totally ignore that property and write a warning log entry

exception is thrown here:
https://github.com/dj-nitehawk/FastEndpoints/blob/f141f7b413b207be176d244814eed4d7b51e2534/Src/Swagger/ValidationProcessor/Extensions/ValidationExtensions.cs#L61

wjax:
Closes #136

### DIFF
--- a/Src/Swagger/ValidationProcessor/Extensions/ValidationExtensions.cs
+++ b/Src/Swagger/ValidationProcessor/Extensions/ValidationExtensions.cs
@@ -46,10 +46,10 @@ internal static class ValidationExtensions
     /// Keys are the property names of the rules, converted to the schema casing by using the selected JsonNamingPolicy
     /// </summary>
     /// <param name="validator"></param>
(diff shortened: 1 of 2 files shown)


dj-nitehawk:
brilliant! works good now...
thanks buddy!

PR #138, opened 2022-05-26
Template pack docs

lazyboy1:
- add basic documentation for template pack. 
- bump template pack version to 1.0.0

Let me know if you think we need to detail more. I think most devs will just look for the installation and basic usage commands.


dj-nitehawk:
thank you very very much 🤩

PR #141, opened 2022-05-26
Fix stackoverflow while generating fluentvalidation swagger data and Limit Validators discovery mechanism to only endpoint validators 

### LINKED ISSUE #139: Stackoverflow in swagger generation when RequestDTO has a circular reference
There is a stackoverflow during the swagger generation when a request dto contains an instance of itself.
The problem happens because the schema processor uses recursion.

The offending request looks like this:
image
or
image

```
  at FastEndpoints.Swagger.ValidationSchemaProcessor.TryApplyValidation(NJsonSchema.JsonSchema, System.Collections.ObjectModel.ReadOnlyDictionary`2<System.String,System.Collections.Generic.List`1<FluentValidation.IValidationRule>>, System.String, System.String)
   at FastEndpoints.Swagger.ValidationSchemaProcessor.ApplyRulesToSchema(NJsonSchema.JsonSchema, System.Collections.ObjectModel.ReadOnlyDictionary`2<System.String,System.Collections.Generic.List`1<FluentValidation.IValidationRule>>, System.String)
```

Reported via Discord at: https://discord.com/channels/933662816458645504/979140098937479258/979140105925177515

### LINKED ISSUE #140: FastEndpoints tries to discover all Validators in a project, even those not related to Endpoints/Requests
Fastendpoints, by default, tries to discover every validator that implements `IValidator` in the whole project.
This behaviour creates problems when user define more than one validator per type (for its own use and not related to the request) as FastEndpoints store them in a dictionary with Key = Type.

Problem lies in this piece of code:
https://github.com/dj-nitehawk/FastEndpoints/blob/2ac2987ff57bca551f03bbd263e122b59d08faf0/Src/Library/Extensions/EndpointData.cs#L116-L121

FastEndpoints should only process and map those validators tied to endpoint requests.

wjax:
Fixes #139 
Fixes #140

### DIFF
--- a/Src/Library/Extensions/EndpointData.cs
+++ b/Src/Library/Extensions/EndpointData.cs
@@ -83,7 +83,7 @@ internal sealed class EndpointData
                     !t.IsGenericType &&
                     t.GetInterfaces().Intersect(new[] {
                         Types.IEndpoint,
-                        Types.IValidator,
+                        Types.IEndpointValidator,
                         Types.IEventHandler,
                         Types.ISummary
                     }).Any());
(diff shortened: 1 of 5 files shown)


wjax:
Sorry to mix two issues in one pull request. They were very simple and felt it was ok.
Just tell me if you prefer two prs

PR #142, opened 2022-05-27
NullReferenceException with MVC APIs on swagger doc

delasource:
In my case, the ApiExplorer lists all API Routes, even classical MVC api routes. Upon loading /swagger/v1/swagger.json page, it throws an exception in the `.Single()` call in OperationSecurityProcessor.cs:26 because those APIs dont have an 'EndpointDefinition'. This PR checks this.

As i was testing my change, i noticed that the `builder.Services.AddSwaggerDoc` method is marked obsolete in the newest NSwag version, so i tried `builder.Services.AddSwaggerDocument` instead - and surprise - it fixed my initial problem.

You may close this PR, but i think, the implemented simple check does not hurt and you may accept it.

### DIFF
--- a/Src/Swagger/OperationSecurityProcessor.cs
+++ b/Src/Swagger/OperationSecurityProcessor.cs
@@ -20,7 +20,7 @@ internal class OperationSecurityProcessor : IOperationProcessor
         if (epMeta is null)
             return true;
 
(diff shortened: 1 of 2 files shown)


delasource:
i investigated further and something is not quite right. i close the PR for now to be safe.

PR #155, opened 2022-06-29
Custom TryParse method should take precedence over JsonSerializer for Enumerable types

fretje:
See also this thread on discord: https://discord.com/channels/933662816458645504/991649707602489367

I think this is a possible fix for that.

Also added a test case for it to the existing RouteValueBinding test, which failed at first (although with a very undescriptive error message - general 500).


dj-nitehawk:
thanks for the fix...

PR #157, opened 2022-07-01
[BUG] fix endpointDiscoveryOptions 

sl-antonybasbug:
EndpointDiscoveryOptions objet is not considered by the action delegate without this fix


dj-nitehawk:
thanks for the fix

PR #174, opened 2022-07-28
Added DoesNotReturn attributes for throwing methods

Maskoe:
[DoesNotReturn Attribute](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.doesnotreturnattribute?view=net-6.0)
This fixes some issues with nullability and analyzers.

image

I ran into it doing this, which should be pretty common.

I couldn't run the tests, because I can't build the project. Something about "DiscoveredTypes" not existing, but I highly doubt this breaks anything.


dj-nitehawk:
thanks!

PR #196, opened 2022-08-18
Overwrite Validation registration

### LINKED ISSUE #190: Outsourcing validators to separate class-lib is not possible
I have the following project setup:
Screenshot from 2022-08-14 15-12-59

My requests, responses and validators are living outside of the API project that references `FastEndpoints` in a project called `Contract`. This is done to allow for easy integration of requests, responses and validators into the API and Blazor frontends.

Due to changes in #141 `AbstractValidator<T>` instances are no longer registered by `FastEndpoints`, so  this doesn't work any longer.
Installing `FastEndpoints` into the `Contract` class lib is not an option as this will break Blazor WASM with the following error:
```
Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(427, 5): [NETSDK1082] There was no runtime pack for Microsoft.AspNetCore.App available for the specified RuntimeIdentifier 'browser-wasm'.
```

MichaelHochriegl:
Allows to override the registration of validators by either:

- set a list of types to be registered as validators during configuration
- set the validator type for an endpoint in the `Configure` method

### DIFF
--- a/Src/Library/Config/EndpointOptions.cs
+++ b/Src/Library/Config/EndpointOptions.cs
@@ -39,4 +39,15 @@ public class EndpointOptions
     /// </code>
     /// </summary>
     public Action<EndpointDefinition>? Configurator { internal get; set; }
+
+    /// <summary>
+    /// set to true if you'd like to define the validator used by an endpoint 
+    /// </summary>
+    /// <code>
+    /// app.UseFastEndpoints(c =>
+    /// {
+    ///    c.Endpoints.SkipValidatorDiscovery = true;
+    /// }
+    /// </code>
+    public bool SkipValidatorDiscovery { internal get; set; }
 }
--- a/Src/Library/Endpoint/Endpoint.Setup.cs
+++ b/Src/Library/Endpoint/Endpoint.Setup.cs
@@ -297,6 +297,13 @@ public abstract partial class Endpoint<TRequest, TResponse> : BaseEndpoint where
     protected void ScopedValidator() => Definition.ScopedValidator();
 
     /// <summary>
+    /// specify the validator that is used for this endpoint.
+    /// if you want to specify the validator for each endpoint you will have to set <see cref="EndpointOptions.SkipValidatorDiscovery"/> to true.
+    /// </summary>
+    /// <typeparam name="TValidator"></typeparam>
+    protected void Validator<TValidator>() => Definition.Validator<TValidator>();
+
+    /// <summary>
     /// specify an override route prefix for this endpoint if a global route prefix is enabled.
     /// this is ignored if a global route prefix is not configured.
     /// global prefix can be ignored by setting <c>string.Empty</c>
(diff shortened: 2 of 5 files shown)


dj-nitehawk:
thanks @MichaelHochriegl 
i will have a look at the code tomorrow.

@wjax if you have some time, would appreciate a code review.

MichaelHochriegl:
There is still the checks missing that you described in #190
I'm currently unsure as to where to throw them.

For example I thought about throwing in the `EndpointDefinition` in the `public void Validator<TValidator>()` method but accessing the `EndpointOptions` via `EpOpts` there they are not yet set (the get set in the `app.UseFastEndpoints` call which is executed after that.

wjax:
Hi there,

I would go with the following:

### Modifications
- Remove `SkipValidatorDiscovery`
    - We already have `CustomValidatorTypes` that can hold all functionality (look at [Usage](#intended-usage))
    - Validator defined in `Configure` method should **take precedence** over any discovered one, and both autodiscovery and manually specified validator (by `Configure` method) **should/could coexists** The PR is actually doing so now.
- Rename `CustomValidatorTypes` --> `AutoDiscoveryValidatorTypes`

### Intended usage 
`AutoDiscoveryValidatorTypes` (former CustomValidatorTypes)  in **EndpointDiscoveryOptions**
 - If null => default behaviour as in PR, use `IEndpointValidator`
 - If empty array => no validator autodiscovery (similar to the removed `SkipValidatorDiscovery`
 - If not empty array => use those and only those types for autodiscovery

Most of the functionality is in the PR and I think it is ok (I have looked at it and do some basic testing).

In case it is needed in the future, I would use FluentValidation to validate Config objects at some point and raise the proper exception

Night

dj-nitehawk:
i'll merge this in to a new branch and have a crack at it today...
i can't get a proper feel of it with the damn diff view 😜

wjax:
You can fetch the automatically created branch of the PR. This is how I reviewed yesterday on local.
And for that Rider is great, love it much more than VS

dj-nitehawk:
i see... still waiting for an OSS license from jetbrains ;-)

do you mean i can add https://github.com/MichaelHochriegl/FastEndpoints/tree/feat/190_validation-registration-config as a remote? or something else? may i pls see a rider screenshot? :-)

wjax:
I think it is something about github. You can checkout the PR as a branch. I follow instructions :) :
image

And then in rider I do the comparison, testing etc:
image

rider_diff

dj-nitehawk:
ah nice...
gitkraken (which i use instead of VS for git stuff) also has it as "add remote and checkout".
but rider side-by-side diff does look pretty descent ;-)
thanks for that 🙏

wjax:
Gitkraken is king, but it is a subscription model and I have already many subscriptions in my life :)
I bought the Rider license last year and I love it, much faster and engineered that VS

PR #228, opened 2022-09-09
Add HTTP 429 and fix typo

MarcusOtter:
I assume this code changes the swagger description. The 429 is common and is used by the library itself with rate limiting, so I think it should be added here.

bild

I also fixed a typo in HTTP 405.

Haven't tested this, but I don't see why it wouldn't work.


dj-nitehawk:
yep looks ok...
thanks!

PR #238, opened 2022-09-15
Query object binding

vpetrusevici:


Src/Library/Binder/RequestBinder.cs
```diff
@@ -149,7 +155,29 @@ private static void BindRouteValues(TRequest req, RouteValueDictionary routeValu
     private static void BindQueryParams(TRequest req, IQueryCollection query, List<ValidationFailure> failures)
     {
```
dj-nitehawk, line 156: this way of repeatedly iterating properties in a nested manner doesn't look that performant to me. since it's gonna happen for every single request, each time. most likely why binding to complex objects from query params like this is not supported in minimal apis. since it's not supported by minimal apis, we don't really need to support it anyway if we can't do it efficiently. also the allocation of dictionaries and json strings and repeatedly calling the json serializer is also not that elegant and will add to the garbage collector pressure.

so going forward, if someone really must do it, they'd either have to add a static `TryParse()` method to the the type or implement a [custom binder](https://fast-endpoints.com/docs/model-binding#diy-request-binding) themselves.

vpetrusevici: Total number of iterations will be the same as for simple query binding.
The main problem here is dictionary parsing. I'll see how we can avoid it. Also I'll try to run some performance test to compare/

Src/Library/Extensions/ReflectionExtensions.cs
```diff
@@ -126,10 +126,23 @@ expression.Body switch
         return Expression.Lambda<Func<object?, (bool, object)>>(
             block,
             inputParameter
-            ).Compile();
+        ).Compile();
     }
 
-    private static object? DeserializeArrayString(object? input, Type tProp)
+    private static object? DeserializeJsonObjectString(object? input, Type tProp)
```
dj-nitehawk, line 132: @vpetrusevici let's just do this and call it a day. this will also work for other binding sources such as form fields, header values, etc.

vpetrusevici: it will work if we send query like ?x={"name":"x","age":24}.
But I want to implement a behavior from mvc [FromQuery]. Query should be ?name=x&age=24

dj-nitehawk: i suggest you implement it in a custom model binder as i feel this is not something i wanna add to the library itself. there must be a really good reason why it's not supported by minimal apis also.

dj-nitehawk:
@vpetrusevici check out [this commit](https://github.com/FastEndpoints/Library/commit/8ea209f5cdbe293be5d8062cd2a3a7857b7bd580)

the following works now.
```cs

// ?id=1&name=john&age=32

public class Request
{
    public int Id { get; set; }

    [FromQueryParams]
    public Person Person { get; set; }
}
```

PR #239, opened 2022-09-15
improve performance of query binding

vpetrusevici:


Src/Library/Binder/RequestBinder.cs
```diff
@@ -155,29 +155,28 @@ private static void BindRouteValues(TRequest req, RouteValueDictionary routeValu
...
-        //        }
-        //    }
-        //    var json = JsonSerializer.Serialize(dictionary);
-        //    var value = JsonSerializer.Deserialize(json, prop.PropType)!;
-        //    prop.PropSetter(req, value);
-        //}
+        var cachedProps = fromQueryProps;
+        foreach (var prop in cachedProps)
+        {
+            var value = prop.ConstructorInfo!.Invoke(Array.Empty<object>());
```
dj-nitehawk, line 161: this is definitely a no no. constructorinfo invoke is very slow. that is why we're using compiled and cached expressions. it's gonna be a nightmare to try and do compiled/cached expressions for all nested property types.

dj-nitehawk:
i think you shouldn't spend any more time on this. whatever solution we come up with is gonna be pretty nasty and it will slow the model binding down to the mvc levels.

Src/Library/Binder/RequestBinder.cs
vpetrusevici: Ok. I'll try to reuse query model with request inheritance then. Thank you for your time!

Issue #218, opened 2022-09-06
Possibility to use 3rd party (non-conforming) DI container


nhaberl:
Hi, I have integrated SimpleInjector into my project and would like to know if fastendpoints is capable of using that instead of buit in container ?

dj-nitehawk:
any third party IOC library compatible with the asp.net middleware pipeline could theoretically be used.
there seems to be some specific way to configure simpleinjector with asp.net according to this: https://docs.simpleinjector.org/en/latest/aspnetintegration.html
give it a shot and post your config here for future readers benefit.

however, also found this gh issue https://github.com/dotnet/aspnetcore/issues/41863 where there seems to be some trouble getting simpleinjector to play nice with the new minimal apis which fastendpoints is built around.

dotnetjunkie:
Would it be possible for FastEndpoints to add an interception point that allows overriding the creation/resolving of endpoints (which seem to be the root components in FastEndpoints)? This would typically mean introducing something like an `IEndpointFactory` class, which in the default implementation forwards the call to MS.DI. For instance:

``` c#
// Interception point
public interface IEndpointFactory
{
    object Create(Type endpointType);
}

//  Default implementation (register as scoped)
public sealed record DefaultEndpointFactory(IServiceProvider Container) : IEndpointFactory
{
    public object Create(Type endpointType) => Container.GetRequiredService(endpointType);
}

// Registration
services.AddScoped<IEndpointFactory, DefaultEndpointFactory>();
```

This allows several integration scenarios such as integration of non-conforming containers (e.g. Simple Injector). For instance, a Simple Injector user could in this case integrate using the following code:

``` c#
// Custom implementation
public sealed record SimpleInjectorEndpointFactory(Container Container) : IEndpointFactory
{
    public object Create(Type endpointType) => Container.GetInstance(endpointType);
}

// Overriding the framework provided implementation
services.AddSingleton<IEndpointFactory>(new SimpleInjectorEndpointFactory(container));
```

Please note that the suggested definition of the `IEndpointFactory` interception point is a guess; I'm unsure whether  this abstraction has the proper signature for both FastEndpoints ánd external parties. It might be useful, for instance, to add extra parameters to the `Create` method, such as an `EndpointContext` of some sort. Although this might not be needed for Simple Injector to work, it could certainly allow other integration scenarios to work.

HTH

dj-nitehawk:
[...]

PR #243, opened 2022-09-18
Adds IEndpointFactory interception point

dotnetjunkie:
This PR is related to the discussion #218 and adds:

* An `IEndpointFactory` abstraction for the creation (or interception) of `BaseEndpoint` instances.
* A `DefaultEndpointFactory` implementation that is used as default implementation of the FastEndpoints framework
* A singleton registration of `DefaultEndpointFactory` in the `AddFastEndpoints` extension method.

The `DefaultEndpointFactory` is both responsible for the created endpoint and the injection of any properties. Property injection should be part of the default implementation, because otherwise it consuming `ExecutorMiddleware` would possibly override any properties that are already injected by the default implementation.

The `DefaultEndpointFactory`, however, is *not* responsible for initialization of the endpoint instance; this is left to the `ExecutorMiddleware`. Letting the factory initialize the endpoint would force the `BaseEndpoint` `Definition` and `HttpContext` property setters to become public; which has possibly bigger design consequences.

The definition of the `IEndpointFactory` is chosen such, that its `Create` method is supplied with as much information as possible. This is why its signature is as follows:

``` c#
BaseEndpoint Create(Endpoint endpoint, HttpContext ctx)
```

By supplying the `Endpoint` instead of a `Type` it gives access to all metadata a custom implementation might want access to. It gives the largest amount of flexibility to someone who wishes to intercept the creation of endpoints.

Supplying `HttpContext` has the same advantages; it allows a custom implementation to base the created endpoint implementation on request-specific information. Another advantage is that it allows the `DefaultEndpointFactory` to be registered as a `Singleton`, as the `DefaultEndpointFactory` now can use the `HttpContext` to resolve request-based services from.

But even though the `DefaultEndpointFactory` is registered as a `Singleton`, the `ExecutorMiddleware` still resolves the `IEndpointFactory` on each request instead of caching it. This is deliberate, because it allows custom `IEndpointFactory` implementations to use another lifestyle. This allows scoped registrations to be injected into its constructor, without having to resolve them explicitly from the `HttpContext`. In other words, this again increases flexibility.

Downside of of letting the `ExecutorMiddleware` resolve the `IEndpointFactory` on each request is the theoretical loss of performance because of the extra invoked resolve. If this proves to be a too costly performance hit (which I actually doubt, as resolving a Singleton from MS.DI is reaally fast), you can also communicate through the `IEndpointFactory` documentation that custom implementations will be treated as `Singleton`. This allows you to cache resolve `IEndpointFactory` instance indefinitely.

Using this PR a Simple Injector user, for instance, would be able to integrate FastEndpoints with Simple Injector as follows:

``` c#
// Custom IEndpointFactory implementation
public record SimpleInjectorEndpointFactory(Container Container) : IEndpointFactory
{
    public BaseEndpoint Create(Endpoint endpoint, HttpContext ctx) =>
        Container.GetInstance(endpoint.Metadata.GetMetadata<EndpointDefinition>()!.EndpointType);
}

// Registration
var builder = WebApplication.CreateBuilder();
builder.Services
    .AddFastEndpoints()
    .AddSimpleInjector(container)
    .AddSingleton<IEndpointFactory, SimpleInjectorEndpointFactory>();
// etc
```

### DIFF
--- a/Src/Library/Extensions/ExecutorMiddleware.cs
+++ b/Src/Library/Extensions/ExecutorMiddleware.cs
@@ -61,6 +61,9 @@ internal class ExecutorMiddleware
 
             BaseEndpoint epInstance = factory.Create(endpoint, ctx);
 
+            epInstance.Definition = epDef;
+            epInstance.HttpContext = ctx;
+
             ResponseCacheExecutor.Execute(ctx, endpoint.Metadata.GetMetadata<ResponseCacheAttribute>());
 
             //terminate middleware here. we're done executing
--- a/Src/Library/Extensions/ExecutorMiddleware.cs
+++ b/Src/Library/Extensions/ExecutorMiddleware.cs
@@ -57,10 +57,9 @@ internal class ExecutorMiddleware
                 }
             }
 
-            var epInstance = (BaseEndpoint)ctx.RequestServices.GetRequiredService(epDef.EndpointType);
-            epInstance.Definition = epDef;
-            epInstance.HttpContext = ctx;
-            ResolveServices(epInstance, ctx.RequestServices, epDef.ServiceBoundEpProps);
+            var factory = ctx.RequestServices.GetRequiredService<IEndpointFactory>();
+
+            BaseEndpoint epInstance = factory.Create(endpoint, ctx);
 
             ResponseCacheExecutor.Execute(ctx, endpoint.Metadata.GetMetadata<ResponseCacheAttribute>());
 
@@ -69,17 +68,6 @@ internal class ExecutorMiddleware
         }
 
         return _next(ctx); //this is not a fastendpoint, let next middleware handle it
-    }
-
-    private static void ResolveServices(object epInstance, IServiceProvider services, ServiceBoundEpProp[]? props)
-    {
-        if (props is null) return;
-
-        for (var i = 0; i < props.Length; i++)
-        {
-            var p = props[i];
-            p.PropSetter(epInstance, services.GetRequiredService(p.PropType));
-        }
     }
 
     private static void ThrowAuthMiddlewareMissing(string epName)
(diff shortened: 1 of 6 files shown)


dj-nitehawk:
give me some days. will review as soon as i get some free time.
thanks!

dotnetjunkie:
I didn't find the websites documentation in the same repository. If you want me to do a PR on the docs, let me know (and point me to the repo containing the docs)

Src/Library/Extensions/ExecutorMiddleware.cs
```diff
@@ -12,10 +12,12 @@ internal class ExecutorMiddleware
 {
     private const string authInvoked = "__AuthorizationMiddlewareWithEndpointInvoked";
     private const string corsInvoked = "__CorsMiddlewareWithEndpointInvoked";
+    private readonly IEndpointFactory _epFactory;
     private readonly RequestDelegate _next;
 
-    public ExecutorMiddleware(RequestDelegate next)
+    public ExecutorMiddleware(IEndpointFactory epFactory, RequestDelegate next)
```
dotnetjunkie, line 18: Do note that since the `ExecutorMiddleware` is effectively a singleton, the developer creating a custom `IEndpointFactory` should also ensure that their implementation runs correctly as a singleton.

dj-nitehawk:
did some benchmarking and decided to cache the factory in the middleware since there seems to be about a 8 to 10% perf drop.

dj-nitehawk:
also changed the signature of the interface since `Endpoint` can be easily obtained from `HttpContext.GetEndpoint()` method.

dj-nitehawk:
if you wanna have a crack at the documentation, it's here: https://github.com/FastEndpoints/Documentation/blob/main/src/routes/docs/%5B...5%5Ddependency-injection.md

do commit to the `dev` branch

Issue #249, opened 2022-09-20
[Feature] Asymmertic JWT Signing


itsmuntadhar:
First of all, I'm a huge fan of this lib and this isn't an issue per se

I was wondering if adding the option to use public/private key pair to sign JWT

One great benefit is other consumers being able to verify the token without going back to the issuer, since they can simply obtain the public key safely and do the verification on their own

dj-nitehawk:
merged. thanks!

PR #250, opened 2022-09-20
JWT asymmetric signing

itsmuntadhar:
Simple solution for #249 

Probably needs more work and tests


dj-nitehawk:
thanks for the PR

PR #253, opened 2022-09-22
Split query params for objects in swagger generation

vpetrusevici:


dj-nitehawk:
why exactly do we need this? i just tried with c# client generation and it looks alright. it generated the following method:
```cs
public virtual async Task<Response> TestGetAsync(string @string, bool @bool, int? @int, long @long, double @double, SystemDayOfWeek @enum, Person person, CancellationToken cancellationToken)
```

vpetrusevici:
For mvc nswag generates the schema as separate query params. Looks like this client generator can handle object, but for example https://github.com/cyclosproject/ng-openapi-gen can't.
Mb this is generator's problem, but there is another one.
I noticed a strange bug for query object with generic type and inheritance like this
```public sealed class GetAdminArticlesPagedRequest
{
    [FromQueryParams]
    public PaginationFilterWithTextAndSort<AdminArticlesSortColumn> Filter { get; init; } = null!;
    public string Locale { get; init; } = null!;
}
public enum AdminArticlesSortColumn : byte
{
    Title,
    Author,
    DateOfCreation
}
public class PaginationFilterWithTextAndSort<T> : PaginationFilterWithTextAndSort where T : struct, Enum
{
    public T SortByColumn { get; init; }
}

public class PaginationFilterWithTextAndSort : PaginationFilterWithText
{
    public bool SortByAscending { get; init; }
}

public class PaginationFilterWithText : PaginationFilter
{
    public string? FilterText { get; init; }
}

public class PaginationFilter
{
    public int PageIndex { get; init; }
    public int PageSize { get; init; }
}
```
And this is what i have in the doc

``` /api/admin/articles/{locale}/paged:
    get:
      tags:
      - GetAdminArticlesPaged
      operationId: GetAdminArticlesPaged
      parameters:
      - name: locale
        in: path
        required: true
        schema:
          type: string
      - name: filter
        in: query
        required: true
        schema:
          $ref: '#/components/schemas/PaginationFilterWithTextAndSortOfAdminArticlesSortColumn'
      responses:
        200:
          description: Success
          content:
            application/json:
              schema:
                type: array
                items:
                  $ref: '#/components/schemas/ArticleAdminPageItemDto'```
```
 PaginationFilterWithTextAndSortOfAdminArticlesSortColumn:
      allOf:
      - $ref: '#/components/schemas/PaginationFilterWithTextAndSort'
      - type: object
        additionalProperties: false
        required:
        - sortByColumn
        properties:
          sortByColumn:
            nullable: false
            $ref: '#/components/schemas/AdminArticlesSortColumn'
    AdminArticlesSortColumn:
      type: integer
      description: ''
      x-enumNames:
      - Title
      - Author
      - DateOfCreation
      enum:
      - 0
      - 1
      - 2
     
    PaginationFilterWithTextAndSort:
      allOf:
      - $ref: '#/components/schemas/PaginationFilterWithText'
      - type: object
        additionalProperties: false
    PaginationFilterWithText:
      allOf:
      - $ref: '#/components/schemas/PaginationFilter'
      - type: object
        additionalProperties: false
    PaginationFilter:
      type: object
      additionalProperties: false
```
Only property from child class was generated for some reason...

vpetrusevici:
Just rechecked it on the example project and can't reproduce... Mb there is some conflict with mvc generation, I'll see

PR #254, opened 2022-09-22
Add support for enums and bools in query object deserialization

vpetrusevici:


dj-nitehawk:
merged...
even though i'm a bit concerned about the recursive reflection use.
would be nice to see a benchmark of mvc controller vs. fast-endpoint ;-)
think you could whip up a benchmark using `benchmarkdotnet`?
if you have some free time that is...

vpetrusevici:
I'll try

PR #257, opened 2022-09-22
Http methods enumes passed by Verbs use bit fields to pass ,

vipwan:
Http methods enums passed by Verbs use bit fields to pass ,  

eg. 

        Routes("OnlyVer3");
        Verbs(Http.POST | Http.GET);

This way of writing should be better concise and interesting


dj-nitehawk:
sorry this is a breaking change and doesn't really add any value. also makes it inconsistent with other methods accepting params array.
😉

PR #259, opened 2022-09-24
Fix null issue when publishing an event and no handlers have been registered

paulczy:


dj-nitehawk:
thank you!
available in `v5.2.0-beta18`

PR #265, opened 2022-09-29
Add benchmark for query binding

vpetrusevici:


vpetrusevici:
Looks like this PR is broken because of some commits in main. I'll convert it to draft for now.

dj-nitehawk:
looks like compared to mvc, your latest binding code is good. just a tiny bit more allocations being made compared to mvc:
|                    Method |     Mean |    Error |   StdDev | Ratio | RatioSD |   Gen0 |   Gen1 | Allocated | Alloc Ratio |
|-------------------------- |---------:|---------:|---------:|------:|--------:|-------:|-------:|----------:|------------:|
| FastEndpointsQueryBinding | 75.58 us | 2.363 us | 1.406 us |  0.85 |    0.02 | 4.0000 | 0.2000 |  32.67 KB |        1.07 |
| AspNetCoreMVCQueryBinding | 88.65 us | 1.657 us | 0.986 us |  1.00 |    0.00 | 3.7000 | 0.2000 |  30.45 KB |        1.00 |

dj-nitehawk:
i removed the minimal api from the bench since it doesn't offer this functionality out of the box and there's no point in adding a tryparse to the model for the sake of benchmarking.

vpetrusevici:
Mb it can be improved using .AsSpan in some places like creating next key

dj-nitehawk:
most likely. gonna have a crack at it?

vpetrusevici:
You can use the same query string for both requests now to make the test more accurate

vpetrusevici:
I can try, but definitely not in the near future. mb on vacation :)

dj-nitehawk:
> You can use the same query string for both requests now to make the test more accurate

actually i think i tried that. but the 3rd level phone numbers didn't bind i think.

PR #266, opened 2022-09-29
Change query params format

vpetrusevici:


dj-nitehawk:
this breaks child binding in swagger ui, because swagger uses the `[...]` brackets for child values
screen-record-1664511072

vpetrusevici:
> this breaks child binding in swagger ui, because swagger uses the `[...]` brackets for child values screen-record-1664511072

I know. I think this is the reason why nswag splitted params. With MVC format we can support object arrays and  have more compatibility.

dj-nitehawk:
we only need to support this format which swagger ui uses for child entities.
and also whatever minimal apis support.
mvc is on it's way out...
so let's stick to the brackets (if both can't be supported easily)?

vpetrusevici:
Minimal api supports the same format as MVC. You can see benchmarks in the second PR

dj-nitehawk:
can we support both? dot and bracket?
if not feasible, priotity should be given to brackets imho to make swagger ui work correctly.
dot is not neccessary.

complex model binding from query parans is a can of worms... 
that is why i didn't wanna touch it ;-)

vpetrusevici:
I think we can. I'll rewrite logic on this weekend

Tests/IntegrationTests/FastEndpoints.IntegrationTests/WebTests/MiscTestCases.cs
```diff
@@ -296,10 +296,10 @@ public async Task BindingFromAttributeUse()
                 "api/test-cases/route-binding-test/something/true/99/483752874564876/2232.12/123.45/" +
                 "?Bool=false&String=everything&XBlank=256" +
                 "&age=45&name=john&id=10c225a6-9195-4596-92f5-c1234cee4de7" +
-                "&numbers=0&numbers=1&numbers=-222&numbers=1000&numbers=22" +
```
dj-nitehawk, line 299: don't change this. we need to support this too.

vpetrusevici: I know. It's still in progress ;)

vpetrusevici:
aaand it's ready to for review.
I didn't fix array of objects yet, but I'll do

Tests/IntegrationTests/FastEndpoints.IntegrationTests/WebTests/MiscTestCases.cs
```diff
@@ -290,6 +290,221 @@ public async Task DupeParamBindingToIEnumerableProps()
...
+                    },
+                    Numbers = new() { 0, 1, -222, 1000, 22 }
+                }
+            }
+            );
+    }
+
+
+    [Fact]
+    public async Task BindingArraysOfObjectsFromQueryUse()
```
dj-nitehawk, line 405: this test is currently not passing. this is still not done right?

vpetrusevici: Yep, I'll try to finish this part today

dj-nitehawk: okie then

vpetrusevici: Mb I'll rewrite some parts to use read-only spans instead of strings

dj-nitehawk:
check this commit: https://github.com/FastEndpoints/FastEndpoints/pull/266/commits/b54193376a1ef4ee5b130f7aff37fccb34d2a428

added some notes.

vpetrusevici:
Can you check it now? I added cache anyway, i think it should be usefull now. All reflexion logic will be exetude only once for each type.
The best way is to use routes trees of ReadonlyMemory, but it takes too much time for me...

dj-nitehawk:
@vpetrusevici i was working on your latest code before your last commit to clean it up a bit.
so now your last commit has some conflicts with what i did.
will you be able to resolve those?

vpetrusevici:
yep, i'll try now

dj-nitehawk:
if it's too messed up to fix. just close this PR and open a new one for your lastest changes against the [FastEndpoints:object-query-binding](https://github.com/FastEndpoints/FastEndpoints/tree/object-query-binding) branch i guess.

vpetrusevici:
I just didn't understand witch difference between 
```i++;
 newRoute = $"{route}[" + i.ToString() + "]";```
and 
```newRoute = $"{route}[{++i}]";```

dj-nitehawk:
> I just didn't understand witch difference between
> 
> ```
> i++;
>  newRoute = $"{route}[" + i.ToString() + "]";
> ```
> 
> and
> 
> ```
> newRoute = $"{route}[{++i}]";
> ```

i use this [vs extension](https://marketplace.visualstudio.com/items?itemName=MukulSabharwal.ClrHeapAllocationAnalyzer) to find closure/heap/boxing allocations.

and it complains that `i` will be boxed/unboxed.

vpetrusevici:
ok then, np

vpetrusevici:
resolved. I can recheck

dj-nitehawk:
good stuff mate!
thank you very much for your effort...
let's see if we get that benchmark going to see if there's any improvements...

vpetrusevici:
at least we already have support for arrays, objects and both query formats)

dj-nitehawk:
yes this is more than good enough for now...
we can slowly optimize from here...

Issue #270, opened 2022-10-04
Validators from `Include()` and `RuleForEach()` are not included into the swagger documentation


laya-alsh:
Hello,

It looks like `ValidationSchemaProcessor` does not respect validators from `Include()` and `RuleForEach()` rules. 

this issue is forwarded from the [small discord conversation](https://discord.com/channels/933662816458645504/1024278454948085780).

I have prepared a small example to demostrate the problem. Imagine you have next request model structure
```c#
public class MyRequest: IRequest
{
    public string? FirstName { get; set; }
    public IReadOnlyList<Child>? Children { get; set; }
    public Partner? Partner { get; set; }
}

public record Child(string? FirstName) : IRequest;

public record Partner(string? FirstName) : IRequest;

public interface IRequest { string? FirstName { get; } }
```

and to validate this request you have next validators
```c#
internal sealed class MyRequestValidator : Validator<MyRequest>
{
    public MyRequestValidator()
    {
        //RuleFor(x => x.FirstName).NotEmpty().MaximumLength(255); // this works
        Include(new BaseValidator()); // this does not work

        RuleFor(x => x.Partner).NotNull().SetValidator(new BaseValidator()); // this works
        RuleForEach(x => x.Children).SetValidator(new BaseValidator()); // this does not work
    }
}

internal sealed class BaseValidator : AbstractValidator<IRequest>
{
    public BaseValidator()
    {
        RuleFor(x => x.FirstName).NotEmpty().MaximumLength(255);
    }
}
```

So for the `MyRequest.FirstName` i would exect in swagger
![image](https://user-images.githubusercontent.com/110024370/193806578-aa4ca846-8a99-485b-b132-98990c64f379.png)

but i got 
![image](https://user-images.githubusercontent.com/110024370/193806433-d983e4ac-f70b-424f-a5d1-b22b635dc046.png)

Similar is for `RuleForEach()`... I noticed that `RuleFor().SetValidator()` works fine, but `RuleForEach().SetValidator()` is ignored..

I've attached a tiny runnable example project.
Thanks!
[validators-bug.zip](https://github.com/FastEndpoints/FastEndpoints/files/9705785/validators-bug.zip)

laya-alsh:
..i'm trying to look into the issue on my own, as the `zymlabs/nswag-fluentvalidation` implementation already has that logic

dj-nitehawk:
@laya-alsh would be nice if you're able to do a PR. i currently have no time to look in to it.

PR #272, opened 2022-10-05
Swagger: apply validation rules from included validators

laya-alsh:
PR for https://github.com/FastEndpoints/FastEndpoints/issues/270

it looks like `RuleForEach()` isn't implemented in the original `zymlabs` repository as well..
but at least `Include()` rules will be applied now

### DIFF
--- a/Src/Swagger/ValidationSchemaProcessor.cs
+++ b/Src/Swagger/ValidationSchemaProcessor.cs
@@ -23,6 +23,7 @@
 using FastEndpoints.Swagger.ValidationProcessor;
 using FastEndpoints.Swagger.ValidationProcessor.Extensions;
 using FluentValidation;
+using FluentValidation.Internal;
 using FluentValidation.Validators;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.Extensions.DependencyInjection;
@@ -100,6 +101,7 @@ public class ValidationSchemaProcessor : ISchemaProcessor
         // Create dict of rules for this validator
         var rulesDict = validator.GetDictionaryOfRules();
         ApplyRulesToSchema(schema, rulesDict, propertyPrefix);
+        ApplyRulesFromIncludedValidators(schema, validator);
     }
 
     private void ApplyRulesToSchema(JsonSchema? schema, ReadOnlyDictionary<string, List<IValidationRule>> rulesDict, string propertyPrefix)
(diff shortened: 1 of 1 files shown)


Src/Swagger/ValidationSchemaProcessor.cs
```diff
@@ -67,15 +68,9 @@ public void Process(SchemaProcessorContext context)
         if (validatorTypes?.Length is null or 0)
             return;
 
-        if (context is null)
-        {
-            _logger?.LogError("SchemaProcessorContext is null");
-            return;
-        }
-
```
laya-alsh, line 75: According to nullable reference type annotations this code is not necessary, so i cleaned up a bit..

Src/Swagger/ValidationSchemaProcessor.cs
```diff
@@ -124,6 +120,37 @@ private void ApplyRulesToSchema(JsonSchema? schema, ReadOnlyDictionary<string, L
...
+            if (adapterMethod.Invoke(adapter, new[] { validationContext, null! }) is not IValidator includeValidator)
+            {
+                break;
+            }
+
+            ApplyRulesToSchema(schema, includeValidator.GetDictionaryOfRules(), string.Empty);
+            ApplyRulesFromIncludedValidators(schema, includeValidator);
+        }
+    }
+
```
laya-alsh, line 153: i pretty much just copied this method from the original `zymlabs` repository

dj-nitehawk:
looks alright at first glance. did you do some testing? 
does it work alright? did it break any existing functionality?
you know, since we don't have any tests for swagger stuff yet.
i'll just have to take your word for it since i don't have much free time to give this a proper vetting ;-)
if working alright, let me know and i'll merge it.

laya-alsh:
yeah, also noticed that there are no tests, otherwise i would have updated/extended them..
i tested it in my project. The only difference in the generated doc (that was a bit not obvious) is
image

but i guess, this change from `OneOf` to `ref`  is because the field is not nullable anymore but a mandatory one

laya-alsh:
oh, you can wait a bit. i found a way to integrate `RuleForEach` rules as well

Src/Swagger/ValidationSchemaProcessor.cs
```diff
@@ -178,23 +178,27 @@ private void ApplyValidationRule(JsonSchema schema, IValidationRule validationRu
...
                 // Retrieve or create an instance of the validator
                 if (!_childAdaptorValidators.TryGetValue(validatorType.FullName!, out var childValidator))
                     childValidator = _childAdaptorValidators[validatorType.FullName!] = (IValidator)Activator.CreateInstance(validatorType)!;
 
                 // Apply the validator to the schema. Again, recursively
-                ApplyValidator(schema.ActualProperties[propertyName].ActualSchema, childValidator, "");
+                var childSchema = schema.ActualProperties[propertyName].ActualSchema;
+                // Check if it is an array (RuleForEach()). In this case we need to apply validator to an Item Schema
+                childSchema = childSchema.Type == JsonObjectType.Array ? childSchema.Item.ActualSchema : childSchema;
+                ApplyValidator(childSchema, childValidator, string.Empty);
```
laya-alsh, line 201: in case of `RuleForEach()` the rules were applied but for a wrong schema... which didn't work obviously..

if you have `RuleForEach(x => x.Children).SetValidator(new ChildValidator())` the validator is for a single `Child` object.. but it was trying here to apply the rules for a `List<Child>` object

laya-alsh:
Now it looks good 🚀

dj-nitehawk:
good job mate!
thank you very much!

dj-nitehawk:
done `v5.2.0-beta24`
should be live in about 5mins...

PR #277, opened 2022-10-13
Skip processors calls for request dto failure

shashikantpawar21:
@dj-nitehawk  
While calling the pre processors I observed it do get the validation failure error, there is no options to separate the errors of pre processor from request dto validation failure... 

These error should be separate from general errors
Also, suppose we use any request dto parameter in preprocessor which is faulty still it will be used in preprocessor and exception error would be thrown 

So, I created this PR .. 
Either we should completely skip the processors in case of validation failure with request object or we should have flag or option for user to throw them separate or add all the errors and throw them together


dj-nitehawk:
it's the responsibility of your pre-processor to check the validity of the input.
this PR cannot be accepted because what if it's a pre-processor that does the logging of validation errors?
so, pre-processors should be executed no matter what.

PR #283, opened 2022-10-16
Remove the read-only properties from the request body in Swagger

WahidBitar:
I've noticed that the swagger example will generate the schema even for the read-only properties which the client will be be able "of course should not be able" to fill out in the request.
So I've just added those read-only properties to the schema.Properties.


dj-nitehawk:
good catch...
thanks for the PR...

PR #286, opened 2022-10-17
additional customized Value Parsers

WahidBitar:
expose additional customized Value Parsers dictionary in the BindingOptions class


dj-nitehawk:
did a bit of refactoring.
this is the end result now:
```cs
app.UseFastEndpoints(c =>
{
    c.Binding.ValueParserFor<Guid>(MyParsers.GuidParser);
});

public static class MyParsers
{
    public static ParseResult GuidParser(object? input)
    {
        Guid result;
        bool success = Guid.TryParse(input?.ToString(), out result);
        return new(success, result);
    }
}
```

Issue #285, opened 2022-10-17
Allow for additional customized Value Parsers


WahidBitar:
I came to a case to use [SmartEnum](https://github.com/ardalis/SmartEnum) in my request models. It was easy to integrate it with the Library binder for the requests that came from the body. However, when the request came from the QueryString I faced a problem.

I know that I can add TryParse to my classes, but I thought if we're able to extend the private dictionary of the parsers to enable the consumers to extend them when registering the endpoints.
e.g.

```
app.UseFastEndpoints(config =>
{
	//....
    config.Binding.CustomParsers.AddSmartEnumParsers();
	//....
}
```
where the `config.Binding.CustomParsers` is `Dictionary<Type, Func<object?, (bool isSuccess, object value)>?>`

WahidBitar:
I've added a PR for this
https://github.com/FastEndpoints/FastEndpoints/pull/286

WahidBitar:
so in case anyone asked how I've registered the SmartEnums here is an example:

```
public static class SmartEnumValueParser
{
    public static void AddSmartEnumParsers(this Dictionary<Type, Func<object?, (bool isSuccess, object value)>?> customParsers)
    {
        var smartEnumTypes = Assembly.GetExecutingAssembly()
            .GetTypes()
            .Where(type => ReflectionUtility.IsDerived(type, typeof(SmartEnum<,>)));

        foreach (var enumType in smartEnumTypes)
        {
            var baseSmartEnumType = ReflectionUtility.GetTypeFromGenericType(enumType, typeof(SmartEnum<,>));
            foreach (MethodInfo methodInfo in baseSmartEnumType.GetMethods())
            {
                if (methodInfo.Name == "FromName")
                {
                    ParameterInfo[] methodsParams = methodInfo.GetParameters();
                    if (methodsParams.Length == 2)
                    {
                        if (methodsParams[0].ParameterType == typeof(string) && methodsParams[1].ParameterType == typeof(bool))
                        {
                            customParsers.Add(enumType, o =>
                            {
                                var enumObj = methodInfo.Invoke(null, new object[] { o.ToString(), true });
                                return new(true, enumObj);
                            });
                        }
                    }
                }
            }
        }
    }
}
```

dj-nitehawk:
looks like an interesting idea. will review the PR sometime this week. thanks!

PR #288, opened 2022-10-19
added a non-generic overload for the ValueParserFor method

WahidBitar:
I noticed that it's easier to add this non-generic overload for the method to do some scenarios to register generic types
like the following example about registering all SmartEnums in the project:

```
app.UseFastEndpoints(config =>
{
	....
    config.Binding.AddSmartEnumParsers();
	...
});
```

```
public static class SmartEnumValueParser
{
    public static void AddSmartEnumParsers(this BindingOptions options)
    {
        var smartEnumTypes = Assembly.GetExecutingAssembly()
            .GetTypes()
            .Where(type => ReflectionUtility.IsDerived(type, typeof(SmartEnum<,>)));

        foreach (var enumType in smartEnumTypes)
        {
            var baseSmartEnumType = ReflectionUtility.GetTypeFromGenericType(enumType, typeof(SmartEnum<,>));
            foreach (MethodInfo methodInfo in baseSmartEnumType.GetMethods())
            {
                if (methodInfo.Name == "FromName")
                {
                    ParameterInfo[] methodsParams = methodInfo.GetParameters();
                    if (methodsParams.Length == 2)
                    {
                        if (methodsParams[0].ParameterType == typeof(string) && methodsParams[1].ParameterType == typeof(bool))
                        {
                            options.ValueParserFor<>()
                            customParsers.Add(enumType, o =>
                            {
                                var enumObj = methodInfo.Invoke(null, new object[] { o.ToString(), true });
                                return new(true, enumObj);
                            });
                        }
                    }
                }
            }
        }
    }
}
```


WahidBitar:
It's related to the following enhancement:
https://github.com/FastEndpoints/FastEndpoints/issues/285

dj-nitehawk:
thank you!

PR #293, opened 2022-10-23
Support For Command Bus Pattern

WahidBitar:
Related to this [discord thread](https://discord.com/channels/933662816458645504/1033637031911379014)
I've added support for the Request/Response pattern.
e.g.

simple class to represent the Request:
```
public class SomeRequest : IRequest<string>
{
    public string SomeDto { get; set; }
}
```
The handler that should handle the request:
```
public class SomeRequestHandler : SomeRequestHandler<SomeRequest, string>
{
    public override Task<string> HandleAsync(SomeRequest requestModel, CancellationToken ct)
    {
        return Task.FromResult("It's OK");
    }
}
```

simply you can create the request and send it to get the response:
`var result = new SomeRequest(){SomeDto = "normal DTO"}.SendAsync();`

### DIFF
--- a/Src/Library/Extensions/EndpointData.cs
+++ b/Src/Library/Extensions/EndpointData.cs
@@ -85,6 +85,7 @@ internal sealed class EndpointData
                      t.GetInterfaces().Intersect(new[] {
                          Types.IEndpoint,
                          Types.IEventHandler,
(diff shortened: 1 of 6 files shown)


WahidBitar:
I'll improve it a bit more

PR #296, opened 2022-10-24
Use the same query for benchmark

vpetrusevici:
```
|                    Method |     Mean |    Error |  StdDev | Ratio | RatioSD |    Gen0 | Allocated | Alloc Ratio |
|-------------------------- |---------:|---------:|--------:|------:|--------:|--------:|----------:|------------:|
| FastEndpointsQueryBinding | 152.2 us |  3.65 us | 1.91 us |  0.86 |    0.05 | 16.9000 |  34.07 KB |        1.10 |
| AspNetCoreMVCQueryBinding | 178.1 us | 14.78 us | 8.80 us |  1.00 |    0.00 | 15.4000 |  31.05 KB |        1.00 |
```


dj-nitehawk:
did you test the result. when i tried the phone numbers didn't bind in the last nested level.
no errors thrown, just when inspecting the result it was null.
can you double check and confirm if it binds correctly?

vpetrusevici:
was a problem in the array strings. For MVC style we should use index [0], [1], ...

PR #301, opened 2022-10-26
Add support for byte arrays in query params

vpetrusevici:
Hi! There is a problem with byte array desirialization in query params, but it should work now.
Is used for sql row version for example.


dj-nitehawk:
great! thanks...

PR #306, opened 2022-10-29
remove readonly properties from open api

ms0815user:
Request properties with private setter were also also listed in the open api. Checking whether the setter method exists and is public fixes this issue.


dj-nitehawk:
thank you!

PR #311, opened 2022-11-02
Public configuration getters

gius:
Make configuration getters and EndpointSummary public to allow custom OperationProcessors based on the original source code.

### DIFF
--- a/Src/Library/Config/EndpointOptions.cs
+++ b/Src/Library/Config/EndpointOptions.cs
@@ -13,7 +13,7 @@ public class EndpointOptions
     /// <summary>
     /// prefix for all routes (example 'api').
     /// </summary>
-    public string? RoutePrefix { internal get; set; }
+    public string? RoutePrefix { get; set; }
 
     /// <summary>
     /// a function to filter out endpoints from auto registration.
--- a/Src/Library/Config/VersioningOptions.cs
+++ b/Src/Library/Config/VersioningOptions.cs
@@ -8,7 +8,7 @@ public class VersioningOptions
     /// <summary>
     /// the prefix used in front of the version (for example 'v' produces 'v{version}').
     /// </summary>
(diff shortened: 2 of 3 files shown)


dj-nitehawk:
this info is already exposed via the `EndpointDefinition` which you can obtain like so:
https://github.com/FastEndpoints/FastEndpoints/blob/60c171af30fcd357cbf06e975d7e290cb92146e3/Src/Swagger/OperationProcessor.cs#L48-L49

use extension method `ctx.GetEndpointDefinition()` which is available in `v5.3.1.3-beta`

gius:
Thank you for your answer. I am sorry, I probably don't get it - let's say you want to copy the OperationProcessor, how do you replace these rows https://github.com/FastEndpoints/FastEndpoints/blob/60c171af30fcd357cbf06e975d7e290cb92146e3/Src/Swagger/OperationProcessor.cs#L58-L59 as `Config.VerOpts.Prefix` and `Config.EpOpts.RoutePrefix` are internal?

The same holds for https://github.com/FastEndpoints/FastEndpoints/blob/60c171af30fcd357cbf06e975d7e290cb92146e3/Src/Swagger/OperationProcessor.cs#L103 and `ProducesResponseTypeMetadata.Example`.

dj-nitehawk:
ah my bad. it is now sorted in `v5.3.1.4-beta`
https://github.com/FastEndpoints/FastEndpoints/blob/d8a4c171cceba2e1541e9622df93720d4a138803/Src/Swagger/OperationProcessor.cs#L58-L59
and
https://github.com/FastEndpoints/FastEndpoints/blob/d8a4c171cceba2e1541e9622df93720d4a138803/Src/Swagger/OperationProcessor.cs#L103

because i'd prefer to keep the config properties setter only.

gius:
Thank you!

PR #318, opened 2022-11-05
Fix json serializer context for query binding

vpetrusevici:


dj-nitehawk:
i removed the numeric parse function dictionary. since we already have this functionality for free from the main parser func dictionary. i.e. less code to maintain.
pls check if i made any mistakes.
let me know if it's ok to merge.

vpetrusevici:
looks good to me 👍🏻

Issue #324, opened 2022-11-07
Caching ExecuteMethod in the HandlerDefinition cause exceptions in the scoped disposable services.


WahidBitar:
Related to this [discord discussion](https://discord.com/channels/933662816458645504/1039086761302818856)

I ran into a problem related to the CommandHandlers
the following will work totally fine at the first shot. 
`def.ExecuteMethod ??= def.HandlerType.HandlerExecutor(tCommand, handler);`
However, if you have any scoped disposable dependencies like DbContext, they will be disposed after the first execution, and because the ExecuteMethod is cached it will try to reuse them again so you'll get an exception.

if we called the `def.HandlerType.HandlerExecutor(tCommand, handler);` on each execution the problem will be solved.

does it harm the performance a lot to call the HandlerExecutor on each execution?

dj-nitehawk:
can you try `v5.3.2.6-beta`
issue is fixed but lost a bit of performance because i couldn't figure out the correct expression tree for the needed logic.
will try again in a couple of days with a fresh mindset.
this will have to do for now.

dj-nitehawk:
#328

PR #325, opened 2022-11-07
prevent caching the HandlerExecutor method

WahidBitar:
prevent caching the HandlerExecutor method to not use the cached resolved services.

related to [this issue](https://github.com/FastEndpoints/FastEndpoints/issues/324)

https://discord.com/channels/933662816458645504/1039086761302818856

### DIFF
--- a/Src/Library/Main/EndpointData.cs
+++ b/Src/Library/Main/EndpointData.cs
@@ -171,7 +171,7 @@ internal sealed class EndpointData
                                              "Consider using Event Pub/Sub pattern instead!");
                     }
 
-                    CommandExtensions.handlerCache.Add(tCommand, new(t));
+                    CommandExtensions.handlerCache.Add(tCommand, t);
 
                     continue;
                 }
--- a/Src/Library/Messaging/Commands/CommandExtensions.cs
+++ b/Src/Library/Messaging/Commands/CommandExtensions.cs
@@ -34,16 +34,15 @@ public static class CommandExtensions
     /// <param name="command">the command to execute</param>
     /// <param name="ct">optional cancellation token</param>
     /// <exception cref="InvalidOperationException">thrown when a handler for the command cannot be instantiated</exception>
-    public static Task ExecuteAsync(this ICommand command, CancellationToken ct = default)
+    public static Task ExecuteAsync<TCommand>(this TCommand command, CancellationToken ct = default) where TCommand : ICommand
     {
         var tCommand = command.GetType();
-
         if (handlerCache.TryGetValue(tCommand, out var handlerType))
         {
             var handler = Config.ServiceResolver.CreateInstance(handlerType);
-            var executeMethod = handlerType.HandlerExecutor(tCommand, handler);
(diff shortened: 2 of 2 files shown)


dj-nitehawk:
we're gonna need to find an alternative because `handlerType.HandlerExecutor(tCommand, handler)` is going to compile the expression on each call. that's why it was cached, so the compilation only happens once.

i'm investigating now. let's see...

WahidBitar:
I've figured it out for the commands without result
`return ((ICommandHandler<TCommand>)handler).ExecuteAsync(command, ct);`
but still have some difficulties with the `ICommand<>`

dj-nitehawk:
i'm trying to get the compiled expression to create an instance of the handler on each execution. something like this but it's currently not working. maybe you can see whats wrong?
```cs
    private static Func<object, CancellationToken, Task<TResult>> HandlerExecutor<TResult>(this Type tHandler, Type tCommand, object handler)
    {
        //object Config.ServiceResolver.CreateInstance(tHandler)

        //Task<TResult> ExecuteAsync((TCommand)cmd, ct);

        var resolverInstance = Expression.Constant(Config.ServiceResolver);
        var typeParam = Expression.Parameter(typeof(Type), "handlerType");
        var serviceProviderParam = Expression.Parameter(typeof(IServiceProvider), "tServiceProvider");
        var createInstanceMethod = typeof(IServiceResolver).GetMethod("CreateInstance", BindingFlags.Instance | BindingFlags.Public | BindingFlags.FlattenHierarchy)!;
        var createInstanceCall = Expression.Call(resolverInstance, createInstanceMethod, typeParam, serviceProviderParam);
        var convertedInstance = Expression.Convert(createInstanceCall, tHandler);

        var execMethod = tHandler.GetMethod("ExecuteAsync", BindingFlags.Instance | BindingFlags.Public | BindingFlags.FlattenHierarchy)!;
        var cmdParam = Expression.Parameter(Types.Object, "cmd");
        var ctParam = Expression.Parameter(typeof(CancellationToken), "ct");
        var methodCall = Expression.Call(convertedInstance, execMethod, Expression.Convert(cmdParam, tCommand), ctParam);

        var xx = Expression.Lambda<Func<object, CancellationToken, Task<TResult>>>(
            methodCall,
            cmdParam,
            ctParam
        );
        var yy = xx.ToString();
        return xx.Compile();

        //.Compile();
    }
```
this bloody expressions api is a bloody nightmare to work with.

WahidBitar:
I'm not totally sure, but I guess it is creating an instance of a handler each time, but the problem is the HandlerExecutor method keeps the reference to the very first instance
```
var handler = Config.ServiceResolver.CreateInstance(def.HandlerType); // this is working right
def.ExecuteMethod ??= def.HandlerType.HandlerExecutor(tCommand, handler); // will compile the executor just for the first time
return ((Func<object, CancellationToken, Task>)def.ExecuteMethod)(command, CT); // but here we're not passing anything related to the new resolved handler
```

Issue #326, opened 2022-11-08
Cannot resolve scoped service 'Web.Services.IEmailService' from root provider.


WahidBitar:
I don't know if it's happening with somebody else, I keep getting
Unhandled exception. System.InvalidOperationException: Cannot resolve scoped service 'Web.Services.IEmailService' from root provider.
at the startup of the sample Web application

`prop.PropSetter(epInstance, ctx.Resolve(prop.PropType));`

dj-nitehawk:
is this happenning within the duration of an http request or outside that duration?

dj-nitehawk:
from the docs:
> If the command handler execution happens outside the duration of an http request and it needs to access a scoped dependency, an exception will be thrown. In such situations you can inject an IServiceScopeFactory in the constructor and create a scope as described in the event handler section above. I.e. the need for a scope factory would only arise if/when you execute commands outside of endpoints.

WahidBitar:
It's related to the Property Injection
there was a public IEmailService property in the Sales.Orders.Create endpoint and the Create endpoint trying to resolve it at the startup, so we're getting this exception.

For now I just removed this public property and will push PR

dj-nitehawk:
all tests are passing. how are you triggering this?

WahidBitar:
I've just tried to run the Web project

dj-nitehawk:
ah right got it... let me investigate...

dj-nitehawk:
problem solved in `5.3.2.11-beta`

PR #327, opened 2022-11-08
Fixed the error of Property Injection from the root service provider

WahidBitar:


WahidBitar:
https://github.com/FastEndpoints/FastEndpoints/issues/326

PR #328, opened 2022-11-08
added CommandHandlerWrapper to resolve the command handlers with TResult

WahidBitar:
I've added some proxy classes to get rid of Dynamic or MethodInfo invoke


dj-nitehawk:
optimized your solution a bit. 
can you check pls.
if everything looks ok now, i can merge.

WahidBitar:
I've checked it out and everything looks ok. I've added a simple endpoint to test the commands from the Web sample page

PR #353, opened 2022-11-28
Accept an optional `CancellationToken` parameter on async map-related methods

danjrwalsh:
Current async versions of `Mapper.FromEntity`, `Mapper.ToEntity`, `Endpoint.MapFromEntity` and `Endpoint.MapToEntity` don't accept a CancellationToken that could be desirable in certain contexts.

As the parameter is optional, it will not cause any breaking changes.


dj-nitehawk:
thank you!

PR #366, opened 2023-01-17
Adds an overload to SendAsync which will accept an object (as long as…

kcadduk:
… it is not of the typed TResponse type).

It will check to see if an IResponseInterceptor has been provided to the Definition of the endpoint, and if it has, run that code.

Then it checks to see if a response has started, if it has, it will simply return, or if not, it will run the SendAsync extension of HttpContext as normal.

Please note: I cannot seem to find the docs to add this information, if you need that too, let me know.

### DIFF
--- a/Src/Library/Endpoint/Auxiliary/EndpointDefinition.cs
+++ b/Src/Library/Endpoint/Auxiliary/EndpointDefinition.cs
@@ -40,9 +40,6 @@ public sealed class EndpointDefinition
     public string? OverriddenRoutePrefix { get; private set; }
     public List<string>? PreBuiltUserPolicies { get; private set; }
     public bool ThrowIfValidationFails { get; private set; } = true;
-    
-    public IResponseInterceptor? ResponseInterceptor { get; set; }
-
 
     //only accessible to internal code
     internal object[]? EpAttributes;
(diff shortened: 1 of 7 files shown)


dj-nitehawk:
looks like we need to call the method `SendInterceptedAsync()` because one of the existing unit tests was failing due to ambiguity between the new send method and the existing generic send method when the endpoints response type is set at `object`.
hope this is ok...

kcadduk:
> looks like we need to call the method `SendInterceptedAsync()` because one of the existing unit tests was failing due to ambiguity between the new send method and the existing generic send method when the endpoint's response type is set as `object`. hope this is ok...

I'm fine with the syntax. One thing that I might change, but i'll wait for your input on this first. Is that if we now have a specific method, i'd like to throw an exception if the interceptor hasn't been set. 

I'm not a fan or arbitrarily blowing up the code when it doesn't need to, but also, I'm not a fan of silently calling SendAsync because they forgot to set an interceptor. Before, interceptors were a part of SendAsync, so an interceptor not being set was kind of okay, but I have a difference of opinion now it has its own SendX terminology. What are your thoughts?

Thanks
Kevin

dj-nitehawk:
yep good idea... go for it...

kcadduk:
Done, I also passed the statusCode into the interface method too. It should have done that from the beginning.

PR #373, opened 2023-02-05
Changed method reference for documentation

dmontgomery:
The documentation pointer in UseFastEndpoints() was self-referencing.  Should have been UseFastEndpointsMiddleware()


dj-nitehawk:
good catch... thanks for the PR...

PR #448, opened 2023-05-20
Add support for properties DI in Unit tests

vpetrusevici:


dj-nitehawk:
@vpetrusevici i did a bit of refactoring. can you check pls? basically, if the DI services are supplied like the following, both property and ctor injection will work:

https://github.com/FastEndpoints/FastEndpoints/blob/14a894eb7052dd9892e49b8fb23e95ff20fed4c9/Tests/UnitTests/FastEndpoints.UnitTests/WebTests/UnitTests.cs#L69-L77

PR #457, opened 2023-06-07
Fix redirect docs

timbze:
301/302 docs was wrong for SendRedirectAsync

[MS source docs](https://learn.microsoft.com/en-us/dotnet/api/Microsoft.AspNetCore.Http.HttpResponse.Redirect?view=aspnetcore-7.0#microsoft-aspnetcore-http-httpresponse-redirect(system-string-system-boolean))


dj-nitehawk:
good catch!
thanks for the PR...

PR #463, opened 2023-07-07
Fix fake event handler test

vpetrusevici:


dj-nitehawk:
wait... whaaaat?
how is this even working?
i didn't know that the IOC container automatically converts a single object registration into an `IEnumerable<T>` 🫤
that's what's happenning here right?

what you register here as a single object:
https://github.com/FastEndpoints/FastEndpoints/blob/f2a05ab55798d266e97ebfc23555f49704dd7a50/Tests/UnitTests/FastEndpoints.UnitTests/EventBusTests.cs#L17

get's injected automatically here right?
https://github.com/FastEndpoints/FastEndpoints/blob/a29e96d931b9958343c502efb4b14bc8e0443531/Src/Library/Messaging/Events/EventBus.cs#L25

dj-nitehawk:
anyway made a commit to hide the `DefaultHttpContext` and use a factory method to instantiate the test event:
https://github.com/FastEndpoints/FastEndpoints/blob/85bc87f5e990adb2ceea7bb72b0cfa180daa79f5/Tests/UnitTests/FastEndpoints.UnitTests/EventBusTests.cs#L19-L22

vpetrusevici:
> wait... whaaaat? how is this even working? i didn't know that the IOC container automatically converts a single object registration into an `IEnumerable<T>` 🫤 that's what's happenning here right?
> 
> what you register here as a single object:
> 
> https://github.com/FastEndpoints/FastEndpoints/blob/f2a05ab55798d266e97ebfc23555f49704dd7a50/Tests/UnitTests/FastEndpoints.UnitTests/EventBusTests.cs#L17
> 
> get's injected automatically here right?
> 
> https://github.com/FastEndpoints/FastEndpoints/blob/a29e96d931b9958343c502efb4b14bc8e0443531/Src/Library/Messaging/Events/EventBus.cs#L25

Yes, you can register many classes/instances of the same abstract type separately and inject them all as IEnumerable<AbstractType>. They all will be added in this collection. It's common in many cases and already used in .net for token claims mappings for example

Issue #471, opened 2023-08-22
Source generator does not discover endpoints in nested classes


ekspedycja:
Hi,

I have been working my project that uses quite a lot of generated code. I am also using vertical-slices approach with class by feature approach. 

Long story short - my use case requires the endpoint class to be a nested class. Currently the code generated by the source generator does not see the class.  

I found a workaround for that if someone is interested.  

This also simplified the code a little bit - BTW is there a reason for so much workarounds and exclusions in current generator's code? 

Here is my approach for file: Src/Generator/EndpointsDiscoveryGenerator.cs

```CS
[Generator(LanguageNames.CSharp)]
public class EndpointsDiscoveryGenerator : IIncrementalGenerator
{
    public void Initialize(IncrementalGeneratorInitializationContext context)
    {
        var typeDeclarationSyntaxProvider = context.SyntaxProvider.CreateSyntaxProvider(
                (sn, _) => sn is TypeDeclarationSyntax,
                (c, _) => Transform(c))
            .Where(p => p is not null);

        var compilationAndClasses = context.CompilationProvider.Combine(typeDeclarationSyntaxProvider.Collect());

        context.RegisterSourceOutput(compilationAndClasses, (spc, source) => Execute(source.Left, source.Right, spc));
    }

    private void Execute(ImmutableArray<ITypeSymbol> typeSymbols, SourceProductionContext spc)
    {
        if (!typeSymbols.Any())
            return;
        var fileContent = GetContent(typeSymbols);
        spc.AddSource(
            "DiscoveredTypes.g.cs",
            SourceText.From(fileContent,
                Encoding.UTF8));
    }

    private ITypeSymbol? Transform(GeneratorSyntaxContext ctx)
    {
        var symbol = ctx.SemanticModel.GetDeclaredSymbol(ctx.Node);
        if (symbol is not ITypeSymbol typeSymbol) return null;
        var isEndpoint = typeSymbol.AllInterfaces.Select(i => new TypeDescription(i)).Intersect(new[]
        {
            new TypeDescription("FastEndpoints.IEndpoint"),
            new TypeDescription("FastEndpoints.IEventHandler"),
            new TypeDescription("FastEndpoints.ICommandHandler"),
            new TypeDescription("FastEndpoints.ISummary"),
        }).Any() && typeSymbol is { IsAbstract: false, DeclaredAccessibility: Accessibility.Public };
        return isEndpoint ? typeSymbol : null;
    }

[...]

PR #472, opened 2023-08-22
#471 Nested classes are not discovered by source generator

ekspedycja:
In current implementation of the Source Generator nested classes are not discovered properly. This may be an issue in case of some of Vertical Slices approaches where feature is a class that consists of multiple subclasses. 

```CS
public static partial class GetAssignedTasks
{
    public record Result()
    {
        public string ResultBlahBlah => "BlahBlahBlah";
    }

    public record Query()
        : TeamRequest<Result>
    {
    }

    public class Validator : AbstractValidator<Query>
    {
        public Validator()
        {
            RuleFor(p => p.GameId).NotEmpty();
            RuleFor(p => p.TeamId).NotEmpty();
            RuleFor(p => p.TeamName).NotEmpty();
        }
    }
    public partial class Endpoint : BaseTeamEndpoint<Query, Result, Result>
    {
        {
            Get("assignedTasks");
        }
    }
}
```

Its not an issue with Endpoint and its inheritance as when class is moved outside the top level class it is discovered properly.

The PR also simplifies the approach with discovering namespaces - IMO current approach is the way around.

### DIFF
--- a/Src/Generator/EndpointsDiscoveryGenerator.cs
+++ b/Src/Generator/EndpointsDiscoveryGenerator.cs
@@ -9,7 +9,7 @@ namespace FastEndpoints.Generator;
 [Generator(LanguageNames.CSharp)]
 public class EndpointsDiscoveryGenerator : IIncrementalGenerator
 {
-     public void Initialize(IncrementalGeneratorInitializationContext context)
+    public void Initialize(IncrementalGeneratorInitializationContext context)
     {
         var typeDeclarationSyntaxProvider = context.SyntaxProvider.CreateSyntaxProvider(
                 (sn, _) => sn is TypeDeclarationSyntax,
--- a/Src/Generator/EndpointsDiscoveryGenerator.cs
+++ b/Src/Generator/EndpointsDiscoveryGenerator.cs
@@ -9,27 +9,17 @@ namespace FastEndpoints.Generator;
 [Generator(LanguageNames.CSharp)]
 public class EndpointsDiscoveryGenerator : IIncrementalGenerator
 {
-    public void Initialize(IncrementalGeneratorInitializationContext context)
+    public void Initialize(IncrementalGeneratorInitializationContext ctx)
     {
-        var typeDeclarationSyntaxProvider = context.SyntaxProvider.CreateSyntaxProvider(
+        var typeDeclarationSyntaxProvider = ctx.SyntaxProvider
+            .CreateSyntaxProvider(
                 (sn, _) => sn is TypeDeclarationSyntax,
                 (c, _) => Transform(c))
             .Where(p => p is not null);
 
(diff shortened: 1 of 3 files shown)


nefarius:
Could you enable allowing edits by maintainers please? Trying to commit a minor style fix, thx.

Src/Generator/EndpointsDiscoveryGenerator.cs
```diff
@@ -9,69 +9,42 @@ namespace FastEndpoints.Generator;
...
-            "Namotion",
-            "StackExchange",
-            "Grpc",
-            "PresentationFramework",
-            "PresentationCore",
-            "WindowsBase"
-    };
-
-    public void Initialize(IncrementalGeneratorInitializationContext context)
+     public void Initialize(IncrementalGeneratorInitializationContext context)
```
nefarius, line 12: Could you eliminate the one redundant space at the start, thanks 😎

nefarius (approved):

dj-nitehawk:
@nefarius 

is this PR ok to be merged other than the space thing?
i didn't look at it properly, but saw that the exclusion list was removed.
is that alright?

nefarius:
The exclusion list got removed; the matching of derived classes is now based on searching for the interfaces including the `FastEndpoints` namespace, I still wanna test it in one of my projects but IMHO unless it is not desired to allow someone to create endpoints in a `Microsoft`, `NuGet` etc. namespace it should work alright.

nefarius:
I have a project where I deliberately had to avoid "NuGet" in my namespaces, this will be perfect to see if I can now do that 😎

dj-nitehawk:
@ekspedycja @nefarius 

did some testing with this today and it was only discovering `public` types.
most of my endpoint classes are `internal sealed`. committed a change that will transform internal types as well.
i'll merge this PR now as it seems to be performing good.

@nefarius if you find any issues in your project, do open up a new issue and we'll look in to it then.

thanks a lot @ekspedycja for this 🙏

nefarius:
> @nefarius if you find any issues in your project, do open up a new issue and we'll look in to it then.

Works!

```cs
namespace FastEndpoints
{
    public static class DiscoveredTypes
    {
        public static readonly global::System.Type[] All = new global::System.Type[]
        {
            typeof(global::NuGetCachingProxy.Endpoints.CatchallEndpoint),
            typeof(global::NuGetCachingProxy.Endpoints.PackageDownloadEndpoint),
        };
    }
}
```

👏

ekspedycja:
I'm glad I was able to help 🙂

One thing came to my mind but I had no chance to test it. Currently the list of available types includes FluetValidation.IValidator. I am not quite sure if this may be an issue if some externals packages implement their own Fluent Validators. In the other hand the generator is run on users code not on imported nugets so this should be fine - or maybe a good idea to test that.

Issue #480, opened 2023-09-06
Job Queues: how to read stored command objects?


YandyZaldivar:
I am using EF as the backend storage for Job Queues.

When you create an entity that implements the `IJobStorageRecord` interface there is not a nice way to deal with the `object` type of the `Command` property.

Everything would be much simpler just by changing the type from `object` to `string`, or add this as an alternative.

When saving to the database it is easy to serialize the object to a json string, but when reading it back you can't deserialize, because you don't know the type of command.

Finding a workaround is possible by storing the full object name and using reflection. But such complexity feels foreign to FastEndpoints.

A nice solution could be to provide an optional function converter like `TCommand From<TCommand>(string value)`.

Or be resilient enough to assume that in the following code `record.Command` could be a json string, and in that case instead of autocasting, you deserialize into a TCommand.

![image](https://github.com/FastEndpoints/FastEndpoints/assets/13922845/ced96f92-4486-4b25-a482-90b5b439d910)

dj-nitehawk:
haven't used EF in almost over 10yrs.
wouldn't something like this work?
```cs
public class MyRecord : IJobStorageRecord
{
    [NotMapped]
    public object Command {
        get {
            return JsonSerializer.Deserialize(CmdString, CmdType);
        }
        set {
            CmdString = JsonSerializer.Serialize(value);
            CmdType = value.GetType();
        }
    }
    public string CmdString { get; set; }
    public Type CmdType { get; set; }

    public string QueueID { get; set; }
    public DateTime ExecuteAfter { get; set; }
    public DateTime ExpireOn { get; set; }
    public bool IsComplete { get; set; }
}
```
didn't try it myself.
if that doesn't work, you'll just have to use a nosql db such as mongodb, litedb, etc.

YandyZaldivar:
Something like that could be the workaround I described, but more complicated. You cannot store the Type CmdType just like that, you need to represent it with a string, and yet more complications will follow, for example, my Job entities are defined in a different assembly than the Commands, so the Entities don't have access to the Command types.

This issue is not related to EF, it will affect other storages, and you are forced to use binary serialization, even with NoSQL, notice the BSON serializer in the MongoDB example:

![image](https://github.com/FastEndpoints/FastEndpoints/assets/13922845/98631609-ba75-458a-bb36-264ee24f1c55)

On top of that, even if a NoSQL db makes sense, why to force ppl into it.

[...]

PR #481, opened 2023-09-07
Allows job queues implementers to customize command parsing

YandyZaldivar:
Being discussed in #480

### DIFF
--- a/Src/Library/Messaging/Jobs/IJobStorageRecord.cs
+++ b/Src/Library/Messaging/Jobs/IJobStorageRecord.cs
@@ -1,4 +1,6 @@
-﻿namespace FastEndpoints;
+﻿using System.Runtime.CompilerServices;
+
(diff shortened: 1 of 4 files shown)


Src/Library/Messaging/Jobs/IJobStorageRecord.cs
```diff
@@ -28,4 +30,10 @@ public interface IJobStorageRecord
     /// indicates whether the job has successfully completed or not.
     /// </summary>
     bool IsComplete { get; set; }
+
+    /// <summary>
+    /// allows implementers to customize command parsing.
+    /// </summary>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
YandyZaldivar, line 37: This attribute is not needed, compilers are very smart, just to express my intention of not disrupting performance even by a function call.

PR #483, opened 2023-09-12
Fix explicit adding an assembly with exclusion name

av0:
* Have an assembly name "SystemFoo" or any other which name starts with a string from the exclusion list
* Add it explicitly `options.Assemblies = new[] { typeof(Program).Assembly }`

**Actual result**
No endpoints discovered

**Expected result**
Endpoints found


dj-nitehawk:
good catch!
thanks for the PR 🙏

PR #488, opened 2023-09-19
adding missing AddAuthorization for benchmarks

cphillips83:
AddAuthorization is no longer called in AddFastEndpoints() and needs to be added to the services for benchmarks.


dj-nitehawk:
good catch!
this is an upcoming minor breaking change: https://github.com/FastEndpoints/FastEndpoints/blob/main/Src/Library/changelog.md
guess i missed updating the benchmark project :-)

Issue #505, opened 2023-10-18
Improve request property validation for nested and collection types within `Endpoint`


murillio4:
Recently I stumbled upon some limitations with validating nested request types within the `Endpoint` class. `FluentValidation` adds the full property chain when adding errors on nested and collection types, the `AddError(Expression, ...)` method in `Endpoint` does not.

Example result:
![Screenshot 2023-10-18 at 18 15 54](https://github.com/FastEndpoints/FastEndpoints/assets/10120072/99b866ff-30bb-4ec7-bf19-7957ecc1b882)

Expected result:
![Screenshot 2023-10-18 at 18 12 23](https://github.com/FastEndpoints/FastEndpoints/assets/10120072/b4d14887-bb99-4654-9e75-6fa31741f27e)

Is this something that you want to support? I have a PR ready, but it may need some performance improvements and cleanup.

PS: Great library, loving it ❤️

PR #506, opened 2023-10-18
Add full member chain when using `Endpoint.AddError(Expression, ...)`

murillio4:
#505

### DIFF
--- a/Src/Library/Extensions/ExpressionHelper.cs
+++ b/Src/Library/Extensions/ExpressionHelper.cs
@@ -46,6 +46,7 @@ internal static class ExpressionHelper
             ConstantExpression ce => ce.Value,
             MemberExpression me => GetValue(me),
             MethodCallExpression mce => GetValue(mce),
+            BinaryExpression be => GetValue(be),
             _ => GetValueCompiled(expression)
         };
 
@@ -58,6 +59,20 @@ internal static class ExpressionHelper
             PropertyInfo pi => pi.GetValue(value),
             _ => throw new NotSupportedException($"[{expression}] is not a supported member expression!")
         };
+    }
+    
+    static object? GetValue(BinaryExpression expression)
+    {
+        if (expression.NodeType != ExpressionType.ArrayIndex)
+            return GetValueCompiled(expression);
+
+        if (GetValue(expression.Left) is not Array array)
+            throw new NullReferenceException($"[{expression.Left}] is null!");
+        
+        if (GetValue(expression.Right) is not int index)
+            throw new NullReferenceException($"[{expression.Right}] is null or unsupported value!");
+        
+        return array.GetValue(index);
     }
     
     static object? GetValue(MethodCallExpression expression)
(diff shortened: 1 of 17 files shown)


murillio4:
Possible path generation alternatives:
[`ExpressionHelper.GetExpressionText`](https://learn.microsoft.com/en-us/dotnet/api/system.web.mvc.expressionhelper.getexpressiontext?view=aspnet-mvc-5.2#system-web-mvc-expressionhelper-getexpressiontext(system-linq-expressions-lambdaexpression))
[`ModelExpressionProvider.GetExpressionText<TModel,TValue>`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.viewfeatures.modelexpressionprovider.getexpressiontext?view=aspnetcore-7.0#microsoft-aspnetcore-mvc-viewfeatures-modelexpressionprovider-getexpressiontext-2(system-linq-expressions-expression((system-func((-0-1))))))

Src/Library/Extensions/ExpressionHelper.cs
```diff
@@ -0,0 +1,43 @@
...
+    private static string FormatIndexerExpression(Expression objectExpression, Expression indexExpression)
+    {
+        if (objectExpression is null)
+            throw new ArgumentNullException(nameof(objectExpression), "Object expression cannot be null.");
+
+        return $"{GetPropertyChain(objectExpression)}[{EvaluateExpression(indexExpression)}]";
+    }
+
+    private static string EvaluateExpression(Expression expression)
+        => Expression.Lambda(expression).Compile().DynamicInvoke() switch
```
dj-nitehawk, line 32: is there really no alternative to recursively compiling the expression on each call? i suppose this would be the biggest perf hit. 

the other two alternatives you linked seems to be from the MVC view stuff, which would have to be added as dependencies to the project. so i don't think we can go that route.

is it not possible to maybe cache the compiled expressions and re-use on subsequent calls?

dj-nitehawk: maybe as a compromise we just output `"ObjectArray[x].Test"` when the index is not a constant?

murillio4: Hmm 🤔 I can do some research and benchmarking of different solutions

Only displaying constant args can be a good compromise

murillio4:
Finding a suitable key for caching the compiled expressions seems to be difficult.

I ended up optimising getting the index value by trying to avoid compiling. Now compiling is only used when expression is eg. `i + 1`

murillio4:
we can also remove compilation as an option. Compilation will not be used in most cases, and there are workarounds.

eg.
`AddError(r => r.type[1+1], "....");` - not allowed
`
var i = 1+1;
AddError(r => r.type[i], "....");
` - allowed

dj-nitehawk:
this should be good enough for now. we'll just add a note to the docs to be careful about a potential perf impact when using non-constant indexers.

thanks a lot for this contribution. good stuff!!!

Issue #508, opened 2023-10-20
Add antiforgery support


vipwan:
https://learn.microsoft.com/zh-cn/aspnet/core/security/anti-request-forgery?view=aspnetcore-6.0

dj-nitehawk:
definitely don't wanna include it be default!
and create a separate PR for this if you want it reviewed.
and please add a integration test to #500 if you want that to be merged.
thanks!

vipwan:
> definitely don't wanna include it by default! and create a separate PR for this if you want it reviewed. and please add a integration test to #500 if you want that to be merged. thanks!

By default, anti is turned off, so there is no performance penalty, but for security reasons such as uploading files, we have to enable it

dj-nitehawk:
data annotations PR has been merged but it does not have the antiforgery code. please submit a separate PR for this.

vipwan:
> data annotations PR has been merged but it does not have the antiforgery code. please submit a separate PR for this.

okay

dj-nitehawk:
@vipwan could you please have a look at [this documentation](https://dev.fastendpoints-doc-site.pages.dev/docs/security#csrf-protection-for-form-submissions-antiforgery-tokens) i wrote for antiforgery? let me know if i got some concepts wrong. thanks!

TijmenUU:
I hope this is the right place to ask a question about the implementation. As far as I can tell the anti forgery configuration is only possible to be set through the `BaseEndpoint.Configure` method that can be overridden? This, in combination with the inability to require it for all endpoints, complicates verifying that all endpoints are safe guarded by anti forgery token validation.

What I like to do with these type of "setup once and forget about it" measures is either force it globally so I hit my head against it when I need an opt out or add a test that checks through reflection that all endpoints have it configured (with possible exceptions manually excluded in the test).

I ended up rolling my own solution with a custom middleware and the use of the `IAntiforgery` interface made available by the `...Services.AddAntiforgery(options =>...` method. I am not sure if that is the intended way of solving this issue?

dj-nitehawk:
@TijmenUU 
you can enable anti-forgery for all form data accepting endpoints with the [global endpoint configurator function](https://fast-endpoints.com/docs/configuration-settings#global-endpoint-options) like so:
```cs
app.UseFastEndpoints(
       c => c.Endpoints.Configurator = ep =>
                                       {
                                           if (ep.FormDataContentType is not null)
                                               ep.EnableAntiforgery();
                                       })
```
[...]

PR #509, opened 2023-10-20
Add antiforgery support

vipwan:
#508


Src/Library/Main/MainExtensions.cs
```diff
@@ -60,6 +63,9 @@ public static IApplicationBuilder UseFastEndpoints(this IApplicationBuilder app,
 
     public static IEndpointRouteBuilder MapFastEndpoints(this IEndpointRouteBuilder app, Action<Config>? configAction = null)
     {
+        //use AntiforgeryMiddleware middleware
+        (app as WebApplication)?.UseMiddleware<Middleware.AntiforgeryMiddleware>();
```
murillio4, line 67: Maybe only use middleware if antiforgery is used?

Could be done by checking if its turned "on" on any endpoint, or by some other configuration

vipwan: > Maybe only use middleware if antiforgery is used?
> 
> Could be done by checking if its turned "on" on any endpoint, or by some other configuration

the Middlewire is really Simple, just when set EndpointDefinition.IsEnlableAntiforgery = true  the validation logic is executed

 (Default:false)  it has no effect  :)

```c#
    public async Task Invoke(HttpContext context)
    {
        //GET请求不需要防伪验证
        if (context.Request.Method == HttpMethods.Get ||
            context.Request.Method == HttpMethods.Trace ||
            context.Request.Method == HttpMethods.Options ||
            context.Request.Method == HttpMethods.Head)
        {
            await _next(context);
            return;
        }
        var endpointDefinition = context.GetEndpoint()?.Metadata.GetMetadata<EndpointDefinition>();
        if (endpointDefinition?.IsEnlableAntiforgery is true)
        {
            try
            {
                await _antiforgery.ValidateRequestAsync(context);
            }
            catch (AntiforgeryValidationException)
            {
                context.Response.StatusCode = StatusCodes.Status400BadRequest;
                await context.Response.WriteAsync("Invalid anti-forgery token");
                return;
            }
        }
        await _next(context);
    }
```

dj-nitehawk: haven't checked the code properly yet. but ideally this middleware should only be added to the pipeline if user explicitly asks to enable it during startup, something like this:
```cs
app.UseFastEndpoints(c => c.Security.EnableAntiForgeryTokens = true);
```
if this value is `false`, endpoint level `EnlableAntiforgery()` call should throw an exception during startup.

@vipwan think you can make that happen?

dj-nitehawk: actually scratch that.... lets make the user do two calls to configure antiforgery.

```cs
builder.Services.AddAntiForgery()

app.UseAntiForgery() // this will be a custom extensionmethod that registers the middleware.
```

and during startup if endpoints say `EnableAntiForgery()` and the user hasn't called `AddAntiForgery()` an exception should be thrown.

how about that?

vipwan: > haven't checked the code properly yet. but ideally this middleware should only be added to the pipeline if user explicitly asks to enable it during startup, something like this:
> 
> ```cs
> app.UseFastEndpoints(c => c.Security.EnableAntiForgeryTokens = true);
> ```
> 
> if this value is `false`, endpoint level `EnlableAntiforgery()` call should throw an exception during startup.
> 
> @vipwan think you can make that happen?

its a good Idea
```c#
app.UseFastEndpoints(c => c.Security.EnableAntiForgeryTokens = true);
```

dj-nitehawk: i think this should be the way:
```cs
builder.Services.AddAntiForgery()

app.UseAntiForgery() // this will be a custom extensionmethod that registers the middleware.
```

dj-nitehawk:
usage changed to this:
```cs
//register services
builder.Services.AddAntiForgery()

//enable middleware
app.UseAntiForgery()

//endpoint config
{
    EnableAntiforgery();
}
```
also changed the 400 response by middleware to use the error response builder method instead of just returning a string message.

dj-nitehawk:
@vipwan is this the correct way to spell your name in english?

https://github.com/FastEndpoints/FastEndpoints/blob/2b4d4ebeef08f603d50b606216a8774c1e3b8b65/Src/Library/changelog.md?plain=1#L73

vipwan:
> @vipwan is this the correct way to spell your name in english?
> 
> https://github.com/FastEndpoints/FastEndpoints/blob/2b4d4ebeef08f603d50b606216a8774c1e3b8b65/Src/Library/changelog.md?plain=1#L73
😃 yes,that's right

PR #518, opened 2023-11-02
Add the ability to handle exceptions from an IPostProcessor

### LINKED ISSUE #517: Enhancement: Update IPostProcessor for Exception Handling
I'd like to propose an enhancement to the `IPostProcessor` interface to include exception handling capabilities. As we continue to expand and refine our system, the ability to gracefully manage and respond to exceptions within post-processing operations becomes increasingly critical.

Currently, post-processors are quite limited as for its capabilities, they are simply pipeline behaviours that runs after an endpoint is done.

I am open to suggestions on the best practices for implementing this and would appreciate any input or discussion on the matter.

tcortega:
Here's a brief example of a possible modification to the IPostProcessor to incorporate exception handling. I've introduced a separate context to manage what was becoming an unwieldy number of parameters.

Additionally, I've transitioned from using a 'for' loop to a 'foreach' loop. Given that it's highly improbable for there to be a registration of thousands of post processors, the negligible performance dip – possibly around 0.000001 milliseconds – isn't a cause for concern.

This proposal is by no means finalized; it's open for further conversation.

Closes #517


tcortega:
This is clearly a breaking change btw.

dj-nitehawk:
made some improvements. could you have a final look to see if everything's alright now?
if all good, we can merge.
thanks!

tcortega:
Hey @dj-nitehawk, everything seems good. Just wondering, what's the reason behind switching the context class into a struct? I'm trying to figure out if there's a benefit I'm missing, because it looks like we might end up with a bunch of unboxing because of the way the Post/Pre processors work.

dj-nitehawk:
dang you're right!
i had my heap allocation viewer plugin turned off 😁
was just looking at the `BinderContext` class and thought processor contexts should follow suit.
fixed now. will merge this.
thanks for the contribution.
much appreciated 🙏🙏🙏

dj-nitehawk:
`v5.19.0.6-beta` pushed to nuget if you wanna test it out in the wild ;-)

Issue #519, opened 2023-11-02
Global error handler reason


IeuanWalker:
Hi, looking at the code for the global error handler, looks like the exception message is always returned - https://github.com/FastEndpoints/FastEndpoints/blob/main/Src/Library/Extensions/ExceptionHandlerExtensions.cs

For security reasons, we generally return a generic message for unhandled exception, unless in debug mode.
Returning the exception message could end up leaking info about the internals of the app that can help with an attack.

Would it be ok if i did a PR that add a parameter to `UseDefaultExceptionHandler` return a generic reason.

Proposed changes is just - 

- New parameter `bool UseGenericReason = false`
- Update the return `Reason = UseGenericReason ? "Something unexpected has happened" : error`

dj-nitehawk:
sounds good. go ahead :-)

dj-nitehawk:
`v5.19.0.4-beta` has been pushed to nuget

PR #520, opened 2023-11-02
New UseGenericReason added to global error handler

IeuanWalker:
Added `UseGenericReason` property to `UseDefaultExceptionHandler`
When set to true will return "Something unexpected has happened"

#519


dj-nitehawk:
thank you very much 🙏

Issue #521, opened 2023-11-02
Improve InternalErrorResponse swagger example


IeuanWalker:
Currently, in the swagger summary i am doing this - 
```csharp
Response<InternalErrorResponse>((int)HttpStatusCode.InternalServerError, "Unhandled error", example: new InternalErrorResponse
{
	Status = "Internal Server Error!",
	Code = 500,
	Reason = "Something unexpected has happened",
	Note = "See application log for stack trace."
});
```

If i add the `DefaultValue` attribute to the `InternalErrorResponse` class properties, i can just do this -
```csharp
Response<InternalErrorResponse>((int)HttpStatusCode.InternalServerError, "Unhandled error");
```

Ok for me to do a PR?

dj-nitehawk:
if doing that, i think we should add the `DefaultValue` attribute to other DTOs such as `ErrorResponse` and `ProblemDetails` to keep the behavior consistent right?

dj-nitehawk:
alright then... go ahead with the PR...

dj-nitehawk:
`v5.19.0.5-beta` pushed to nuget.
thanks again ;-)

PR #522, opened 2023-11-02
DTO default values

IeuanWalker:
Added default values where i can.

2 places i couldn't add a default value, as the attribute parm needs to be a constant -

- ErrorResponse.cs `public Dictionary<string, List<string>> Errors { get; set; } = new();`
- ProblemDetails.cs `public IEnumerable<Error> Errors { get; private set; }`

#521


IeuanWalker:
will look at fixing the tests now

PR #523, opened 2023-11-02
Add GlobalPre/PostProcessors and integration/unit test amendment

rudiv:
This adds a  `GlobalPreProcessor` and `GlobalPostProcessor` class to support state-driven global processors.

The existing unit/integration tests were hijacked as the change is relatively straightforward and related, though they could be separated if needed?

### DIFF
--- a/Src/Library/Endpoint/Processor/GlobalPostProcessor.cs
+++ b/Src/Library/Endpoint/Processor/GlobalPostProcessor.cs
@@ -4,6 +4,22 @@
 /// inherit this class to create a global post-processor with access to the common processor state of the endpoint
 /// </summary>
 /// <typeparam name="TState">type of the common processor state</typeparam>
-public abstract class GlobalPostProcessor<TState> : PostProcessor<object, TState, object>, IGlobalPostProcessor where TState : class, new()
+public abstract class GlobalPostProcessor<TState> : IGlobalPostProcessor where TState : class, new()
 {
+    /// <summary>
+    /// not intended for direct external use.
+    /// </summary>
+    [HideFromDocs]
+    public Task PostProcessAsync(IPostProcessorContext context, CancellationToken ct)
+        => PostProcessAsync(context, context.HttpContext.ProcessorState<TState>(), ct);
+
+    // ReSharper disable once MemberCanBeProtected.Global
+    /// <summary>
+    /// implement this method to define the post-processing logic using the provided context and state.
+    /// </summary>
+    /// <param name="context">the context object encapsulating all necessary information for post-processing.</param>
+    /// <param name="state">the common processor state object, derived from the HttpContext or newly instantiated.</param>
+    /// <param name="ct">cancellation token.</param>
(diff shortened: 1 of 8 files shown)


dj-nitehawk:
#518 was merged to master and it seems to have broken your PR.
bunch of tests are failing.
would you be able to sort that out pls?

rudiv:
Tests are fixed, but I'm more curious as to why it didn't (after fixing it to just inherit from `PostProcessor<object, TState>`).

I know #518 was marked as a breaking change but it's a potentially serious breaking change for anyone using `object` as the `TRequest` as that doesn't seem to work any more. Instead, the typeless `IPre/PostProcessor` has to be used.

rudiv:
Hmmm yes this is the underlying error now:

> Unable to cast object of type 'FastEndpoints.PreProcessorContext`1[FastEndpoints.EmptyRequest]' to type 'FastEndpoints.IPreProcessorContext`1[System.Object]'.

I believe it's due to the fact that `EmptyRequest` is a struct. Changing it to a class makes everything pass again. Given that's the case, I'm not sure why this is only coming up after the merge.

dj-nitehawk:
just tried this PR branch with the following and things seem to be working ok:
```cs
app.UseFastEndpoints(
       x => x.Endpoints.Configurator = ep =>
       {
           ep.PreProcessors(Order.Before, new GlobalPreProc());
       });
       
class GlobalPreProc : GlobalPreProcessor<ProcState>
{
    public override Task PreProcessAsync(IPreProcessorContext ctx, ProcState state, CancellationToken ct)
    {
        state.Id = 100;

        return Task.CompletedTask;
    }
}

class ProcState
{
    public int Id { get; set; }
}

sealed class MyEndpoint : Endpoint<EmptyRequest>
{
    {
        Get("test");
        AllowAnonymous();
    }

    public override async Task HandleAsync(EmptyRequest r, CancellationToken c)
    {
        var procState = ProcessorState<ProcState>();
        await SendAsync(procState.Id);
    }
}
```

i'm not seeing any invalid cast exceptions. can you update the above code and send back to replicate what you're seeing?

cheers!

rudiv:
No, everything is fine in what I pushed. I added the methods in to the abstract class (rather than inheriting from `PostProcessor<object, TState>` as it did before).

Previously, this was possible (modifying your code):
```cs
class GlobalPreProc : PreProcessor<object, ProcState>, IGlobalPreProcessor
{
    public override Task PreProcessAsync(IPreProcessorContext<object> ctx, ProcState state, CancellationToken ct)
    {
        state.Id = 100;

        return Task.CompletedTask;
    }
}
```

After the change, it's not and you will get the cast error.

dj-nitehawk:
ah got it...
so i believe this is ready to be merged then?
looks all good to my eyes.

rudiv:
Good for me!

dj-nitehawk:
done!
appreciate the contribution 🙏🙏🙏

PR #524, opened 2023-11-03
Avoid creating string if doing structured logging

IeuanWalker:
Minor change, the `msg` variable is only needed when not doing structured logging.
So this just avoids creating a new string variable


dj-nitehawk:
nice one...
thanks!

PR #527, opened 2023-11-04
New security options, new overloads for AddJWTBearerAuth and sensible defaults

### LINKED ISSUE #526: [Proposal] Alter default JWT Bearer Options
When using `AddJWTBearerAuth` from FE, it's almost guaranteed that the consumer is using `JWTBearer.CreateToken`. It's then reasonable to expect that what you create is what you'll get back, however ASP.NET Core's default behaviour is to map inbound claims.

For example, a `sub` claim will become `http://schemas.xmlsoap.org/ws/2005/05/identity/claims/nameidentifier`.

A test case is available here: https://gist.github.com/rudiv/ee386f4532fc8fa94932a939893dd819

This is fixed in 1 of 2 ways.

1. Globally remove the inbound claim mappings using `JwtSecurityTokenHandler.DefaultInboundClaimTypeMap.Clear();`. This is advice from [this article about OpenId Connect](https://learn.microsoft.com/en-us/aspnet/core/security/authentication/claims?view=aspnetcore-8.0#claims-namespaces-default-namespaces).

There are issues with this however. 1: it doesn't work with .NET 8 (preview 7 and above). 2: It GLOBALLY disrupts the claim mapping, meaning for other schemes (think Identity/Cookies if used, not the FE ones) claims will not map correctly without also updating the Identity/UserManager configuration (as some internals are using `ClaimTypes.NameIdentifier` in the case of the "sub" claim, for example).

For 1, the new code is actually `JsonWebTokenHandler.DefaultInboundClaimTypeMap.Clear()` as a new handler has been added as of preview 7. 2 still applies, however. Note that the link above, despite being 8.0 documentation does not reflect the change yet (a separate issue was logged for this).

2. Use the more modern method of setting `MapInboundClaims = false` on the `JWTBearerOptions`.

This is available since .NET 5 and works by setting the instance property on the handler of the specific authentication scheme.

This proposal thus proposes 1 of 2 options.

1; Expose the full `JWTBearerOptions`, rather than the more inner properties of `TokenValidationParameters` and `JwtBearerEvents`.

This means that the API would change from:
```cs
.AddJWTBearerAuth(Keys.SignKey,
        tokenValidation: o => o.ValidateIssuer = true,
        bearerEvents: o => o.OnTokenValidated = context => ..)
```

to this:
```cs
.AddJWTBearerAuth(Keys.SignKey, options: o =>
{
    o.TokenValidationParameters.ValidateIssuer = true;
    o.Events.OnTokenValidated = context => ..;

    o.MapInboundClaims = false; // Added to disable the mapping
});
```

The benefits of this is less maintenance on the API in case something else was added to `JwtBearerOptions` that might be relevant to developers. Drawback is that it would change the signature of the method and thus be a breaking change (unless it was done as an overload).

2. Within `AddJWTBearerAuth`, set the `MapInboundClaims = false` directly thus making it the default behaviour to not map.

Whilst not adding as much flexibility as the first option, it does provide the expected behaviour to developers that are creating their own signed tokens within the app and expect them to be consistent.

-----

A bonus 3rd option would be to do both, provide a sensible default for the FE specific scheme that could be overridden by the action provided by option 1.

rudiv:
Fixes #526 

This:
- Adds new overloads for the configuration of JWTBearerAuth, including the option to configure the whole `JWTBearerOptions`
- Sets sensible defaults for the JWTBearerAuth, specifically:
- - No longer maps inbound claims
- - Overrides the Role claim to `role` (`Conf.SecOpts.RoleClaimType`)
- - Overrides the Name claim to `name` (`Conf.SecOpts.NameClaimType`)
- Adds new security options (`Conf.SecOpts`) for the Role and Name claim so that it can be customised back to the crazy SOAP ones by doing `Conf.SecOpts.RoleClaimType = ClaimTypes.Role` for example
- `JWTBearer.CreateToken` is using the `Conf.SecOpts.RoleClaimType`

All of the above allows for smaller tokens to be generated and consumed, without unexpected behaviour, whilst still being fully customisable.

Note that, due to the change in the claim role type this is a "breaking change", though the impact should be minimal and will only affect older tokens.

### DIFF
--- a/Src/Directory.Build.props
+++ b/Src/Directory.Build.props
@@ -1,7 +1,7 @@
 <Project>
     <PropertyGroup>
 
-        <Version>5.19.0.9-beta</Version>
+        <Version>5.19.0.10-beta</Version>
 
         <TargetFrameworks>net6.0;net7.0;net8.0</TargetFrameworks>
         <LangVersion>latest</LangVersion>
--- a/Src/Directory.Build.props
+++ b/Src/Directory.Build.props
@@ -1,7 +1,7 @@
 <Project>
     <PropertyGroup>
 
-        <Version>5.19.0.8-beta</Version>
+        <Version>5.19.0.9-beta</Version>
 
         <TargetFrameworks>net6.0;net7.0;net8.0</TargetFrameworks>
         <LangVersion>latest</LangVersion>
--- a/Src/Directory.Build.props
+++ b/Src/Directory.Build.props
@@ -1,7 +1,7 @@
 <Project>
     <PropertyGroup>
 
-        <Version>5.19.0.7-beta</Version>
+        <Version>5.19.0.8-beta</Version>
 
         <TargetFrameworks>net6.0;net7.0;net8.0</TargetFrameworks>
         <LangVersion>latest</LangVersion>
--- a/Src/Library/Config/SecurityOptions.cs
+++ b/Src/Library/Config/SecurityOptions.cs
@@ -6,6 +6,15 @@
 public sealed class SecurityOptions
 {
     /// <summary>
+    /// specify a custom claim type used to identity the name of a user principal. defaults to `name`.
+    /// <para>
+    /// WARNING: do not change the default unless you fully comprehend what you're doing!!!
+    /// </para>
+    /// </summary>
+    public string NameClaimType { internal get; set; }
+        = "name";
+    
+    /// <summary>
     /// specify a custom claim type used to identify permissions of a user principal. defaults to `permission`.
     /// <para>
     /// WARNING: do not change the default unless you fully comprehend what you're doing!!!
@@ -13,4 +22,13 @@ public sealed class SecurityOptions
     /// </summary>
     public string PermissionsClaimType { internal get; set; }
         = "permissions"; //should never change from "permissions" or third party auth providers such as Auth0 won't work.
+    
+    /// <summary>
+    /// specify a custom claim type used to identify roles of a user principal. defaults to `role`.
+    /// <para>
+    /// WARNING: do not change the default unless you fully comprehend what you're doing!!!
+    /// </para>
+    /// </summary>
+    public string RoleClaimType { internal get; set; }
+        = "role";
 }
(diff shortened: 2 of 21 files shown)


rudiv:
While we're at it, super fun bonus - I've updated `JWTBearer.CreateToken()` to use the new `JsonWebTokenHandler` on .NET 8+ (since that's a compile target now too), so less allocations + more performance. These are preliminary benchmarks basically because my computer's deciding to be a bit rubbish and make .NET 8 slower overall. I'll re-run them in a bit.

| Method                | Runtime  | Mean     | Error     | StdDev    | Ratio | RatioSD | Gen0    | Gen1   | Allocated | Alloc Ratio |
|---------------------- |--------- |---------:|----------:|----------:|------:|--------:|--------:|-------:|----------:|------------:|
| FastEndpoints         | .NET 6.0 | 22.61 us |  8.841 us |  5.261 us |  1.00 |    0.00 |  5.1000 |      - |  10.38 KB |        1.00 |
| 'Token (Old Handler)' | .NET 6.0 | 44.14 us |  1.223 us |  0.640 us |  1.98 |    0.39 | 15.2000 |      - |  30.47 KB |        2.93 |
|                       |          |          |           |           |       |         |         |        |           |             |
| FastEndpoints         | .NET 7.0 | 20.40 us | 11.973 us |  7.920 us |  1.00 |    0.00 |  1.6000 |      - |  10.22 KB |        1.00 |
| 'Token (Old Handler)' | .NET 7.0 | 32.68 us |  2.353 us |  1.231 us |  1.63 |    0.52 |  4.9000 | 0.2000 |  29.89 KB |        2.93 |
|                       |          |          |           |           |       |         |         |        |           |             |
| FastEndpoints         | .NET 8.0 | 29.91 us | 21.522 us | 14.235 us |  1.00 |    0.00 |  1.6000 |      - |  10.08 KB |        1.00 |
| 'Token (Old Handler)' | .NET 8.0 | 29.20 us | 13.522 us |  7.072 us |  0.96 |    0.46 |  4.5000 | 0.2000 |  27.35 KB |        2.71 |
| 'Token (New Handler)' | .NET 8.0 | 36.08 us | 26.100 us | 17.264 us |  1.33 |    0.36 |  2.8000 | 0.1000 |  17.53 KB |        1.74 |

KateyBee:
Merged and pushed build out to Nuget `v5.19.0.10-beta`

Did a bit of cleaning up and there were a couple of ambiguous method warnings in `AuthExtensions.cs`.

Let me know if you see anything wrong.

Thanks for the contribution!

Issue #150, opened 2022-06-08
Swagger generation does not honor Minimum length rule after a NotEmpty rule


wjax:
As seen here:
![image](https://user-images.githubusercontent.com/309918/172565550-ee60fe9f-6798-4e20-af01-083ec3238573.png)
![image](https://user-images.githubusercontent.com/309918/172565587-0f1024df-37a2-45de-a47f-cbf5fa9500c4.png)
![image](https://user-images.githubusercontent.com/309918/172565629-88951049-0249-4325-9a20-baac40781859.png)

Link to discord message: 
https://discord.com/channels/933662816458645504/983483088401215488/983483095305060372

PR #579, opened 2024-01-25
Fix minLength swagger generation #150

emrekara37:
Fix: #150


dj-nitehawk:
thanks for the fix. much appreciated 🙏🙏🙏

PR #601, opened 2024-02-07
Copy ContentType value from IFormFile's to StreamContents in test helper

poneymusical:


dj-nitehawk:
thank you 👌

poneymusical:
Thanks for the great library, it’s been a blast using  it! Do you have any idea when this will get into the nuget feed?

dj-nitehawk:
just pushed `v5.22.0.1-beta` to nuget.
you can use that until the next release which would be happenning in the 1st week of march.

PR #105, opened 2022-05-04
Added structured logging to the ExceptionHandler

Toyotomi-clan:
Currently the exception handler does not log structurally.
image

The changes to this pull request allows structured output for the exceptions captured in this logger. 

Here is an example of the stracktrace path 

image

Here is an example of the exception path

image

Both cases are structured logged, enabling tools like [serilog.exceptions](https://github.com/RehanSaeed/Serilog.Exceptions) to output clear errors. 

Or at least see and enable user to search against the exceptions using params i.e. stacktrace or exception depending on the path instead of a blob of text.

PR #625, opened 2024-02-28
Pass handled exceptions to logger

mattwcole:
Exceptions handled in `ExceptionHandlerExtensions` are not currently passed to `ILogger.LogError()`. This is required for logger providers to correctly extract structured data.

#105 tried to fix this in the wrong way IMO. The PR mentions [Serilog.Exceptions](https://github.com/RehanSaeed/Serilog.Exceptions?tab=readme-ov-file#what-does-it-do) which shows passing an exception to the logger in its first example.

The default console logger prints the exception after the message, which is very similar to the current output:

```diff
 =================================
 HTTP: GET /api/things
 TYPE: MyException
 REASON: My exception message.
 ---------------------------------
+MyException: My exception message.
    at ...
    at ...
```

### DIFF
--- a/Src/Library/Extensions/ExceptionHandlerExtensions.cs
+++ b/Src/Library/Extensions/ExceptionHandlerExtensions.cs
@@ -49,13 +49,13 @@ public static class ExceptionHandlerExtensions
                             else
                             {
                                 logger.LogError(
+                                    exHandlerFeature.Error,
                                     $"""
                                      =================================
                                      {route}
                                      TYPE: {exceptionType}
                                      REASON: {reason}
                                      ---------------------------------
-                                     {exHandlerFeature.Error.StackTrace}
                                      """);
                             }
 
--- a/Src/Library/Extensions/ExceptionHandlerExtensions.cs
+++ b/Src/Library/Extensions/ExceptionHandlerExtensions.cs
@@ -45,17 +45,18 @@ public static class ExceptionHandlerExtensions
                             var reason = exHandlerFeature.Error.Message;
 
                             if (logStructuredException)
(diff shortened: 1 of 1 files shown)


dj-nitehawk:
the `else` branch you modified is only meant for textual logging and not for structured logging.
check [my commit](https://github.com/FastEndpoints/FastEndpoints/pull/625/commits/122d430577974fdede2d324898456413484559a5) which i believe will now allow structured logging if you specify it like so:
```cs
.UseDefaultExceptionHandler(logStructuredException: true)
```

mattwcole:
That will do the job thanks! I'd still consider passing the exception to the logger when the message template is unstructured since this is always expected when logging an exception. Users might want to search for exceptions in their log aggregation without the addition of `@exceptionType', `@route` and '@reason' properties.

dj-nitehawk:
if we do that, the there's a duplication of info for example in the console log:
```
fail: FastEndpoints.ExceptionHandler[0]
      =================================
      HTTP: GET test
      TYPE: InvalidOperationException <<<<<
      REASON: wklgjhrkljwhr <<<<<
      ---------------------------------
      System.InvalidOperationException: wklgjhrkljwhr <<<<<
         at MyEndpoint.HandleAsync(CancellationToken c) in D:\SOURCE-CONTROL\Playground\Source\Program.cs:line 28
         at FastEndpoints.Endpoint`2.ExecAsync(CancellationToken ct) in D:\SOURCE-CONTROL\FastEndpoints\Src\Library\Endpoint\Endpoint.cs:line 98
         at FastEndpoints.Endpoint`2.ExecAsync(CancellationToken ct) in D:\SOURCE-CONTROL\FastEndpoints\Src\Library\Endpoint\Endpoint.cs:line 130
```
people including myself, don't want to see that duplication in the console/ or any other text logs.

PR #648, opened 2024-03-20
Fill route parameters in test url

### LINKED ISSUE #573: Integration tests with path parameters
Say I have an endpoint with the given url:

`Get("/api/users/{UserId}");`

and the request dto:

```c#
record RequestDto(string UserId);
```

If I call a `GETAsync` method on an endpoint in an integration test I get the following request url generated:

`"/api/users/{UserId}"`

I have a middleware that relies on getting a UserId value from a path like so:

`var userId = HttpContext.Request.RouteValues["userId"]`

which obviously fails because there is not a valid id in the url.

Perhaps I am doing this the wrong way? Or is this scenario currently unsupported and I should specify URL explicitly?

alexalok:
Fixes #573 

Just a proof of concept for now.

### DIFF
--- a/Src/Library/Testing/HttpClientExtensions.cs
+++ b/Src/Library/Testing/HttpClientExtensions.cs
@@ -1,4 +1,4 @@
-﻿// ReSharper disable InconsistentNaming
+// ReSharper disable InconsistentNaming
 
 using System.Net.Http.Json;
 using System.Reflection;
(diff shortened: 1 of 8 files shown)


alexalok:
@dj-nitehawk what do you think is the right approach regarding route parameters with a value of null? Currently I am just skipping over such parameters, leaving a `/{param}/` as-is. Another option would be replacing with `/null/`.

dj-nitehawk:
did a bit of cleanup & formatting.
looks alright to my eyes now.
you can finish up the other methods + add an integration test or two and it should be good to be merged.

alexalok:
I think everything is ready for a final review. I have updated the logic to ignore props that we know can't be used to hydrate url (marked by either FromClaim, FromHeader or HasPermission). I have also created an integration test for all supported verbs.
As a note, I have slightly refactored HttpClientExtensions to minimize code repetitions at the expense of a possible extra nesting call. If this is undesired, I can revert the changes.

dj-nitehawk:
okay... i believe that should do it.
good job mate!
will merge now 🙌

dj-nitehawk:
`v5.23.0.12-beta` has been pushed to nuget if you wanna give it a whirl.
thanks again for this PR.

alexalok:
Thank you for a quick review! I will def try out the beta in integration tests of one of my projects.

PR #664, opened 2024-04-23
Added so we can set the signing algoritm for the refresh token service

oll1w0ll1:
Had some issues since the release where the default algorithm for the JwtCreationOptions became HmacSha256Signature since a lot of non .NET validators doesn't support this. Added the option so that we can set that in the TokenService constructor. If there's already a way to achieve this, please let me know 😃.


dj-nitehawk:
i've consolidated the two algo properties into one.
could you please give it a test run and see if everything's in order?
if so, we can merge it.
thanks!

oll1w0ll1:
Looks good! And worked as intended 👍 Thank you.

dj-nitehawk:
`v5.24.0.8-beta` has been pushed to nuget with the change.

PR #678, opened 2024-05-09
Add Null-conditional operator to Client.Dispose in the DisposeAsync method

kylelaverty:
I noticed while trying to write some tests that I got an error from the dispose async method. When I looked it was not null checking to prevent calling dispose on a null object.

This should address that issue.


dj-nitehawk:
merged...
thank you!

PR #684, opened 2024-05-15
Fixed ACL generator incorrectly including other props besides the permissions in the permission lists

### LINKED ISSUE #683: Bug: Generated ACL incorrectly contains the descriptions dictionary in the permission dictionarys
# Description

When using the source generator for ACL the `public static Dictionary<string, string> Descriptions ...` also gets included in `_permNames` and `__permCodes` resulting in incorrect values being returned by the utility functions like `AllCodes()`.

# Steps to reproduce

- Scaffold a new FE project using one of the official templates
- Install the FastEndpoints.Generator package
- Go into `SayHello.Endpoint` and register a permission in the following way:
```
    {
        Post("/api/hello");
        AllowAnonymous();
        
        AccessControl("TestPermission");
    }
```
- Now either use the debugger to call a utility function on the `Allow` class, or use the logger like so:
```
public override async Task HandleAsync(Request r, CancellationToken c)
    {
        Logger.LogInformation(JsonSerializer.Serialize(Allow.AllCodes()));
...
```
- The logged codes should look similar: `["System.Collections.Generic.Dictionary\u00602[System.String,System.String]","PDW"]
`

Everything described here was using the latest version on FE (5.25.0) and .net 8.0.

# Temporary workaround for those facing this

Extend the partial allow class like so:
```
public partial class Allow
{
    static partial void Describe()
    {
        // Fixed the source generator incorrectly registering the Descriptions prop as a permission
        _permCodes.Remove("System.Collections.Generic.Dictionary`2[System.String,System.String]");
        _permNames.Remove("Descriptions");
    }
}
```

Xeretis:
Fixes #683

### DIFF
--- a/Src/Generator/AccessControlGenerator.cs
+++ b/Src/Generator/AccessControlGenerator.cs
@@ -200,9 +200,12 @@ public class AccessControlGenerator : IIncrementalGenerator
                  {
                      foreach (var f in typeof(Allow).GetFields(BindingFlags.Public | BindingFlags.Static))
                      {
-                         var val = f.GetValue(null)?.ToString() ?? string.Empty;
-                         _permNames[f.Name] = val;
-                         _permCodes[val] = f.Name;
+                         var val = f.GetValue(null) as string;
+                         if (val != null)
+                         {
+                             _permNames[f.Name] = val;
+                             _permCodes[val] = f.Name;
+                         }
                      }
                      Groups();
                      Describe();
--- a/Src/Generator/AccessControlGenerator.cs
+++ b/Src/Generator/AccessControlGenerator.cs
@@ -156,6 +156,7 @@ public class AccessControlGenerator : IIncrementalGenerator
                 """
 
                 #region DESCRIPTIONS
+                    [HideFromDocs]
                     public static Dictionary<string, string> Descriptions = new()
                     {
                 """);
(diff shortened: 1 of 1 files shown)


Src/Generator/AccessControlGenerator.cs
```diff
@@ -200,9 +200,12 @@ static Allow()
                  {
                      foreach (var f in typeof(Allow).GetFields(BindingFlags.Public | BindingFlags.Static))
```
dj-nitehawk, line 201: wouldn't just filtering out the `Descriptions` dictionary from here be a better alternative?
i.e. the generated `Allow.g.cs` has the following:
```cs
#region DESCRIPTIONS
    public static Dictionary<string, string> Descriptions = new()
    {
        //Customers_Create
        { "JHH", "Permission for creating a new customer in the system." },
        //Sales_Order_Create
        { "LYO", "Allows creation of orders by anybody who has this permission code." },
    };
#endregion
```
which is where the rogue `"System.Collections.Generic.Dictionary`2[System.String,System.String]"` entry comes from. so i think just filtering out this field by name `Descriptions` should be the way to go.

what do you think?

Xeretis: Yeah that would work just fine. The reason I chose to filter strings in general to be included is so if in the future some other similar functionality gets added to the genrator then this issue wouldn't resurface. But then again it could also cause issues in the future that don't exist now... so it's hard to decide. Now I think either soultion is fine.

Xeretis: Or if someone choses to add something similar to the class by extending it. I haven't check yet wheter that would cause the same tho.

dj-nitehawk: we can decorate the `Descriptions` property with the existing attribute `[HideFromDocs]` and use that to exclude the field. if we filter by attribute, users can also use the same strategy (decorate with [HideFromDocs]) if you they wanna add some public static fields/properties to their partial class and it will be automatically excluded. sounds like a plan?

Xeretis: Yeah that sounds great. I can make the changes in about an hour.

Xeretis: Alright , thanks for the suggestions.

Xeretis: Pushed a new commit, that should do it.

dj-nitehawk:
thank you very much 🙏

PR #693, opened 2024-05-23
Fix params parsing when path is using Google rest api guidelines

idormenco:
After upgrading to 5.25 our API crashed during startup. 
After investigation problem was detected in `OperationProcessor.ParamCreationContext`
Fixed it to allow routes named using google rest api guidelines
`/api/{id}:deactivate`


KateyBee:
Thanks for the PR, greatly appreciate the addition of tests :-)
But, the regex didn't work correctly for route patterns with multiple constraints such as this:
```cs
/api/{id:min(1):max(5)}:deactivate
```
So I got rid of the regex and added another Where() clause to filter candidates.
Seems to be working correctly now.
Could you give it another look and if everything seems ok, I will merge.

Thanks!

idormenco:
@KateyBee sorry for that miss, honestly i didn't even knew that that type of constraint is even possible 😺

dj-nitehawk:
looks good to me.
thanks @idormenco for the PR 🙏

PR #735, opened 2024-07-11
FIX: param creation bug for api/a:{id1}:{id2} type routes

ridvandev:
This pull request aims to fix a bug in route creation for patterns like api/a:{id1}:{id2}.


KateyBee:
@ridvandev Thanks for the PR 🙏

Issue #734, opened 2024-07-10
Bug: swagger gen crashes when static serilog configuration is used


vzam:
Version: 5.27.0.6-beta
Version without the same issue: 5.27.0

I have tried the fix for #733 in the new beta and since I am using Serilog in my project, I have discovered that it breaks after the upgrade from 5.27.0. The issue is that I am using `CreateBootstrapLogger` from Serilog, which freezes the logger at some point and apparently FE builds the service provider at some point internally, causing an exception in Serilog due to the freeze.

Here is a minimal example to reproduce:

```csharp
using FastEndpoints;
using FastEndpoints.Swagger;
using Serilog;

var logger = new LoggerConfiguration()
    .WriteTo.Console()
    .CreateBootstrapLogger();

Log.Logger = logger;

var builder = WebApplication.CreateBuilder(args);

builder.Services
    .AddFastEndpoints() // todo: also add some endpoint to avoid an exception here
    .SwaggerDocument();

builder.Host.UseSerilog((_, config) => config.WriteTo.Console());

var app = builder.Build();

app
    .UseFastEndpoints()
    .UseSwaggerGen()
    .UseSwaggerUi()
    ;

app.Run();
```

dj-nitehawk:
should be sorted in `v5.27.0.9-beta` courtesy of @nhart12

PR #736, opened 2024-07-12
Using an injected ServiceProvider instead of building manually

nhart12:
Fixes https://github.com/FastEndpoints/FastEndpoints/issues/734


dj-nitehawk:
good job!
didn't know about the overload with the service provider.
thanks!

PR #772, opened 2024-09-21
Switched to .NET 9.0 System.Threading.Lock

MarkCiliaVincenti:


dj-nitehawk:
thanks for the PR but...
you're bringing in a third party dependency `Backport.System.Threading.Lock` to add support for pre .net 9 sdks yes?
we definitely don't wanna do that for something that doesn't offer much of a benefit.
if you want, you can just do conditional compilation blocks to support the new `Lock` type on .net 9 onwards.

MarkCiliaVincenti:
> thanks for the PR but... you're bringing in a third party dependency `Backport.System.Threading.Lock` to add support for pre .net 9 sdks yes? we definitely don't wanna do that for something that doesn't offer much of a benefit. if you want, you can just do conditional compilation blocks to support the new `Lock` type on .net 9 onwards.

Yes, that's what the micro dependency basically does, but also allows you to use the full functionality of the Lock class.

Src/Messaging/Messaging.Remote/Server/Events/EventHub.cs
```diff
@@ -44,7 +44,14 @@ sealed class EventHub<TEvent, TStorageRecord, TStorageProvider> : EventHubBase,
     static readonly Type _tEvent = typeof(TEvent);
     static bool _isRoundRobinMode;
     static TStorageProvider? _storage;
-    static readonly Lock _lock = new();
+
+    readonly
```
MarkCiliaVincenti, line 48: This is no longer static

dj-nitehawk: oops...
good catch!
fixed.

MarkCiliaVincenti:
@dj-nitehawk a new version of Backport.System.Threading.Lock has come out that acts as a source generator and basically dropping the DLL as a dependency. If this new development makes you reconsider, I will amend accordingly.

dj-nitehawk:
will check what the source generator creates and get back to you on that.

MarkCiliaVincenti:
Handled via #824

PR #778, opened 2024-09-27
Allow HttpClientExtensions.ToForm to have nullable forms

Thijs5:
I have a use case where a file is submitted together with other form data. The file itself is optional. Updated the test extension method so it can handle null files.


dj-nitehawk:
thank you very much!

PR #785, opened 2024-10-04
Centralize package version management

Lenny32:
Updated multiple .csproj files to remove specific version numbers from PackageReference elements, allowing projects to use the latest package versions by default. Reformatted XML for consistency and readability. Introduced Directory.Packages.props for centralized package version management, ensuring consistent versions across the solution. Updated FastEndpoints.sln to include the new solution item and corrected project type GUID.


dj-nitehawk:
thank you so much...
been meaning to do this but never got around to it.
greatly appreciated 🙏🙏🙏

PR #786, opened 2024-10-04
.Add route syntax coloration

Lenny32:
Add StringSyntax attribute for routes in .NET 7.0+

Introduce conditional compilation to include System.Diagnostics.CodeAnalysis
namespace and apply StringSyntax("Route") attribute to route-related parameters
in HttpDeleteAttribute.cs, HttpGetAttribute.cs, HttpPatchAttribute.cs,
HttpPostAttribute.cs, HttpPutAttribute.cs, Endpoint.Setup.cs, and Endpoint.cs
when targeting .NET 7.0 or greater.


dj-nitehawk:
do you know if `StringSyntax` attribute is taken into consideration by both visual studio and rider, or is it only VS that respects it? 

if both IDEs respect `StringSyntax`, can we remove the rider related `[RouteTemplate]`, do you think?

Lenny32:
@dj-nitehawk I do not have a license for Rider unfortunately and I do not know if it works or not. 
`RouteTemplate` is not working on Visual Studio.

dj-nitehawk:
`5.30.0.3-beta` pushed to nuget.

dj-nitehawk:
okay so...

the `[StringSyntax]` attribute doesn't seem to have any effect in the `FastEndpoints.Attributes` package because the NET7 conditional compilation doesn't work because that project is a `netstandard2` project. we don't want to change that either because it's netstandard because this pkg is being used in blazor wasm by some people.

screen-record-1728187136

i'll be stripping out `[StringSyntax]` from `FastEndpoints.Attributes` project because it has no effect unfortunately :-(

PR #805, opened 2024-10-29
Add overload to SendErrorAsync

ali-asi:
Added overload to SendErrorsAsync with HttpStatusCode


dj-nitehawk:
FE doesn't use the `HttpStatusCode` enum for status codes. instead uses `int` status codes everywhere. we don't like the enum due to it being too verbose for conveying an intention that's usually just 3 digits to read, and devs should be fully familiar with the actual status codes instead of them being hidden behind enum names. if you really must use the enum, it's easy to write your own extension methods targetting the `IEndpoint` interface.

image

Issue #151, opened 2022-06-12
Cannot describe nested objects in request with RequestParam method. 


DCichosz:
![image](https://user-images.githubusercontent.com/40369900/173250345-09965509-4e37-45a9-aaf7-1922eb61ed11.png)

Arcalise08:
@dj-nitehawk I can tackle this one if nobody else is working on it.  It correlates fairly well with the work mentioned in #566 

Which is taking a little longer then expected but it's been one fire after another so far this year 😅

dj-nitehawk:
yes please go ahead 🙏
and take your time. this one's pretty low priority.
cheers!

sunecko:
Hello, I opened a pr I hope and it meets expectations. I remain on the lookout for any change request or whether it meets the requirements or not. I'm excited to finally be able to contribute to my favorite .NET library.

dj-nitehawk:
@sunecko thank you for the contribution. but i think you've misunderstood the requirement here.
take the following example:
```cs
sealed class Request
{
    public int Id { get; set; }
    public NestedRequest Nested { get; set; }
}

sealed class NestedRequest
{
    public string Name { get; set; }
}

sealed class Summary : Summary<MyEndpoint, Request>
{
    public Summary()
    {
        RequestParam(r => r.Id, "this shows up in swagger doc");
        RequestParam(r => r.Nested.Name, "this should appear in swagger doc, but it doesn't");
    }
}

sealed class MyEndpoint : Endpoint<Request>
{
    public override void Configure()
    {
        Post("test");
        AllowAnonymous();
    }

    public override async Task HandleAsync(Request r, CancellationToken c) { }
}
```
which produces swagger schema like this:
![image](https://github.com/user-attachments/assets/df439512-fd66-43bb-8658-9eda075cc1cf)

the text description for the nested property is missing. that is the feature that's currently not supported. only the top level property descriptions can be provided like above atm.

sunecko:
my mistake, not all the changes were added correctly to the commit, the main change is here
```c#
public class EndpointSummary<TRequest> : EndpointSummary where TRequest : notnull
{
    /// <summary>
    /// add a description for a request param for a given property of the request dto
    /// </summary>
    /// <param name="property">a member expression for specifying which property the description is for</param>
    /// <param name="description">the description text</param>
    public void RequestParam(Expression<Func<TRequest, object?>> property, string description)
    {
        var propertyPath = GetPropertyPath(property);
[...]

PR #808, opened 2024-11-04
Add support for nested objects in RequestParam

sunecko:
#151


dj-nitehawk:
i tested with [this example](https://github.com/FastEndpoints/FastEndpoints/issues/151#issuecomment-2454940999) and it doesn't seem to work.

can you check with that example request structure pls?

this is what i see:

image

the description for `Nested > Name`  property is missing.

sunecko:
WhatsApp Image 2024-11-14 at 11 27 22_a149e6b6

dj-nitehawk:
i added support for collections but it only works with array and list properties. should be good enough for now i think.
give it a test run and let me know if you see any problems.
if it looks alright, will merge.

Issue #821, opened 2024-11-15
Endpoint version vs Release version


P1Rh0:
I'm having an issue with versioning. I want to have a well defined overall release version at each stage, defined in Swagger with some subset of various endpoint versions.
I can easily version each endpoint using the `Version` configuration option and then set `MaxEndpointVersion` to define various swagger versions. The issue I have is that with the following three releases:

Release 0:
_/endpointA_
_/endpointB_

MaxEndpointVersion: 0 in swagger

Release 1:
_/endpointA/v1_
_/endpointB_

MaxEndpointVersion: 1 in swagger

Release 2:
_/endpointA/v1_
_/endpointB/v1_

MaxEndpointVersion: 1 in swagger?

When I define release 2, release 1 will actually use the v1 version of _/endpointB_

With that not working, I looked at using the EndpointFilter option instead but realised that
1. I still have to use an arbitrary large number for `MaxEndpointVersion`
2. I now have to tag every endpoint with every version. I can initially tag everything with "RV1" but then have to retag for every release which seems like a pain

It would be useful if we could add a release version to the versioning at the same time. In the example above this would look like:

Release 0:
_/endpointA_ - Version(0, releaseVersion: 0) 
_/endpointB_ - Version(0, releaseVersion: 0)

MaxReleaseVersion: 0 in swagger

Release 1:
_/endpointA/v1_ - Version(1, releaseVersion: 1) 
_/endpointB_ - Version(0, releaseVersion: 0)

MaxReleaseVersion: 1 in swagger

Release 2:
_/endpointA/v1_ - Version(1, releaseVersion: 1) 
_/endpointB/v1_ - Version(1, releaseVersion: 2)

MaxReleaseVersion: 2 in swagger

dj-nitehawk:
will look in to this...
give me a few days pls...

P1Rh0:
No worries, I am happy to help/spend time on this if it's something you agree is needed. There's a PR open but what I will need is guidance and a bit of domain knowledge :) Thanks for looking into it!

PR #822, opened 2024-11-15
Add release versioning

P1Rh0:
A stab at resolving #821. Not sure if this is everything required. Especially not sure about the change around tagSegments in DocumentProcessor.cs

### DIFF
--- a/Src/Library/Endpoint/Auxiliary/EndpointDefinition.cs
+++ b/Src/Library/Endpoint/Auxiliary/EndpointDefinition.cs
@@ -601,6 +601,7 @@ public sealed class EndpointDefinition(Type endpointType, Type requestDtoType, T
 public sealed class EpVersion
 {
     public int Current { get; internal set; }
+    public int ReleaseVersion { get; internal set; }
     public int DeprecatedAt { get; internal set; }
 
     internal void Init()
--- a/Src/Swagger/DocumentOptions.cs
+++ b/Src/Swagger/DocumentOptions.cs
@@ -57,6 +57,11 @@ public class DocumentOptions
     public int MinEndpointVersion { get; set; }
 
     /// <summary>
+    /// endpoints with a release version greater than this version will not be included in the swagger doc.
+    /// </summary>
+    public int MaxReleaseVersion { get; set; }
+
+    /// <summary>
     /// by default deprecated endpoints/operations will not show up in the swagger doc.
     /// set this to true if you instead want them to show up but displayed as "obsolete".
     /// </summary>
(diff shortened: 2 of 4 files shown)


dj-nitehawk:
so i've gone ahead and implemented a revised (hopefully better) approach to release versioning.
have a look at the [sandbox project's program.cs file](https://github.com/P1Rh0/FastEndpoints/blob/main/Sandbox/Source/Program.cs) for an example of this in action.

basically, the idea is:

- when defining a swagger document, you give the doc a release version number.
- endpoints can then associate themselves with a particular swagger doc and upwards just by doing the following:

```cs
Versiong(n).StartingRelease(2);
```

that endpoint will only show up under swagger docs with a release version number 2 or higher.
seems to solve the original conundrum you mentioned in #821 

please give it a good test and let me know if you see any issues.
if everything seems alright, we can merge this PR into the `main` branch.

btw, had to do this method chaining approach to prevent breaking changes with the current versioning strategy.

P1Rh0:
Ah amazing :) Thanks so much and sorry I was a bit wide of the mark with my implementation.

Testing now and seeing an error when opening swagger:
`System.Collections.Generic.KeyNotFoundException: The given key '__isLatest__' was not present in the dictionary.`

Looking into it now but you might know what it is straight away so thought I'd drop a comment here too

dj-nitehawk:
the dictionary lookup issue should be sorted now. if not, let me know how to reproduce it.
and, no worries about the implementation. took me nearly 2 days of trying different things to arrive at this.

P1Rh0:
Yep that now runs. 

- [x] Versions greater than the release version work fine, e.g. `Version(2).StartingRelease(1)`.
- [x] Later release versions also seem to correctly pick up the latest version, e.g. `Version(1).StartingRelease(1)` shows up in `o.ReleaseVersion = 2;`
- [ ] Versions smaller than the release version _do not show_, e.g. `Version(1).StartingRelease(2)` doesn't show up in `o.ReleaseVersion = 2;`

In that final case I would expect to see the endpoint, is that something you're able to look into?

dj-nitehawk:
can you check again pls? i see the following:

image

which matches the criteria: `Version(1).StartingRelease(2)`

if that is not the endpoint you're talking about, let me know which one. or just update the [sandbox project's program.cs file](https://github.com/P1Rh0/FastEndpoints/blob/main/Sandbox/Source/Program.cs) to show only the scenario you're talking about.

thanks!

IsaacClem:
Hmm yeah that does look like a me problem, I'll take another look and get back to you, thanks :)

P1Rh0:
That does look like a me problem! Let me take a look and get back to you, thanks :)

P1Rh0:
Sorry, work got busy! 
Tested against the lastest release this morning and I think Ive identified the issue. I have a GET and a DELETE endpoint with the same route (`/entity/{entityId}`), in the case above I only get the DELETE showing in swagger.
I'll try to update the sandbox with an example

P1Rh0:
Done :) Example screenshot below (probably depends on whichever it finds first as in this case its the DELETE that is removed not the GET)
image

dj-nitehawk:
thanks for that repro...
fixed in `v5.32.0.1-beta`

P1Rh0:
Loooks good to me :) Thanks!

PR #824, opened 2024-11-18
Cleaned up .NET 9.0 locking via source generator.

MarkCiliaVincenti:


dj-nitehawk:
sorry mate!
decided to go with the type alias instead of brining in even a source generator.
simply because we're only using a lock in just 2 places and none of the advance usage coming from the new lock object is going to be used for the foreseeable future.
might look at your library in the future if we do.
cheers!

PR #841, opened 2024-12-06
Request mapper can have update entity

AlexAlappsis:


dj-nitehawk:
do you think someone would ever want an async variant of this new method as well?

AlexAlappsis:
Fair question, I don't know.  I saw that ToEntity and FromEntity have async versions, but wasn't sure when I would use them, so it didn't occur to me to add it.  I was just copying the implementation from the Mapper class, which does not having an async of UpdateEntity.  

I suppose it wouldn't hurt to add to both Mapper and RequestMapper, to have parity with the To and From functions.

dj-nitehawk:
yeah, go ahead and add it.
will merge then.

PR #848, opened 2024-12-12
Fix multipart form data when it consists for JSON objects

eshmanev:
I have recently faced a problem POSTing multipart form data using client.POSTAsync<TEndpoint, TRequest, TResponse>(...).
The problem is caused by the .ToString() call that just saves the object type name instead of serializing it. Therefore, it's better to call JsonSerializer.Serialize() instead of .ToString(), so that complex values can be handled properly.  

How to reproduce:
1. Prepare a request containing complex values, for example an array of other objects
2. Send the request using POSTAsync

Actual result: A validation error happens "MyType[] is not valid for a property MyType[]"
Expected result: No validation errors

### DIFF
--- a/Src/Library/Testing/HttpClientExtensions.cs
+++ b/Src/Library/Testing/HttpClientExtensions.cs
@@ -422,7 +422,11 @@ public static class HttpClientExtensions
                     AddFileToForm(file, p);
             }
             else
-                form.Add(new StringContent(p.GetValue(req)?.ToString() ?? ""), p.Name);
+            {
+                var value = p.GetValue(req);
+                var stringValue = value is not null ? JsonSerializer.Serialize(value) : null;
+                form.Add(new StringContent(stringValue ?? ""), p.Name);
+            }
         }
 
         return !form.Any()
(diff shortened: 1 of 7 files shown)


dj-nitehawk:
added a test and did a minor refactor to get the test pass.
could you have a look and see if everything works as expected now?
if ok, will merge.
thanks!

eshmanev:
@dj-nitehawk nice improvement. Looks great.
Thanks!

dj-nitehawk:
merged.
thank you for the contribution.

SpaceOgre:
@dj-nitehawk  Not sure if I should post here or create a new issue, but I have a problem with submitting data as form and found this PR and tried out the beta version it is in. I think it solved my issues with nullable Guids being set to Guid.Empty instead of null but now I get these errors for Guid with values:

`{"name":"legalProvisionId","reason":"Value [\"8267237c-a8f4-4882-bc05-4e1b7d665093\"] is not valid for a [Guid] property!"}`

dj-nitehawk:
@SpaceOgre i think it would be best to create a new issue with how exactly to reproduce the problem. a debuggable repro project attached to the issue would be much appreciated. will have a look at it as soon as i can...

eshmanev:
@dj-nitehawk when are you going to release this change?

dj-nitehawk:
it's in v5.33 released today

PR #872, opened 2025-01-24
Allow job executions also be cancelled by app shutdown

SjorsGoering:
I noticed that in the JobQueue class in the ExecuteAsync call of a command it passes a CancellationToken which only triggers by time. However I think you also want to be able to cancel any work in the command execution when the application shuts down. Made a small change to make this possible, let me know @dj-nitehawk what you think.


dj-nitehawk:
looks good...
merged...
appreciate the improvements 🙏🙏🙏

PR #884, opened 2025-02-14
Allow combining both IHasTrackingID and IHasCommandType

SjorsGoering:
Made a small change so that commands can implement both the IHasTrackingID as well as IHasCommandType interface.


dj-nitehawk:
good catch! thank you!

PR #885, opened 2025-02-14
Only retry EventHub init Tasks that failed

SjorsGoering:
Made some changes to only retry the failed EventHub initializations.
And some extra additions:
- Logging via source generator
- Added check if it is an in-memory storage provider and skipped initialization, since in that case it makes no sense
- Combined the 30 sec timeout CancellationToken together with the app shutdown CancellationToken
- Added an virtual OnRestoreSubscriberIDsError (for consistency with other places in the class)


dj-nitehawk:
thanks for this...
i cleaned up a bit.

PR #895, opened 2025-02-23
Align various nuget package versions and add net6.0 to FE.AspVersioning

SjorsGoering:
Various nuget package versions have been aligned to the target framework, except for Microsoft.Extensions.Logging because at least v7.0.0 is required due to the use of the logging source generator and presumably the usage of the global using C# feature.

Also added net6.0 support for FastEndpoints.AspVersioning.

I'm wondering if the netstandard target frameworks are still needed (since the docs state net6.0+), so I've left those for now. Let me know your thoughts @dj-nitehawk


dj-nitehawk:
**Asp.Versioning.Http**
---
they only added minimal api support after v7 of their lib.
this is what happens if i target net6.0 with your change: https://streamable.com/cq30ae
so we can't support net6.0 for `FastEndpoints.AspVersioning`
and FE will stop supporting `net6.0` soon because it's no longer supported by microsoft.

**Microsoft.Extensions.DependencyInjection.Abstractions**
---
v8.0.2 is the latest build that net6.0, net7.0 and net8.0 can use.
so what we had in the `Directory.Packages.props` was fine.
image
likewise, v9.0.2 supports both net8.0 and net9.0
same story applies to `Microsoft.Extensions.Hosting` and `Microsoft.Extensions.Primitives`

**FastEndpoints.Messaging.Remote.Core**
---
this lib is supposed to be able to run in the browser with blazor-wasm projects, and blazor-wasm only supported `netstandard2.1` if i remember correctly. so it makes sense just to leave `netstandard2.1` there and not add any other TFMs. if a consumer is building against net6+, the `netstandard2.1` will work just fine.

so... i don't think this PR can be accepted.
thanks for looking in to it though 🙏

dj-nitehawk:
> Maybe this has never been issue for anybody yet

yes that is correct. will look into this again if somebody reports a problem with this approach.

cheers!

PR #897, opened 2025-02-25
Added closed generics support for command jobs

SjorsGoering:
Currently the UseJobQueues registration ignores registered command types that are generic (both open and closed generics), because it wants to create a JobQueue<,,,> instance using the MakeGenericType method which does not support open generics type parameters. Actually the generic type check is too broad, since closed generic types are supported with the MakeGenericType method.

Making the generics check more specific to only allow closed generics, allows an user to register a closed generic command and use it as a job like this:

 ```cs
// Definitions
public class OrderCreatedEvent : IEvent;

public class QueueCommand<TEvent> : ICommand where TEvent : IEvent
{
    public required TEvent Event { get; set; }
}

public class QueueCommandHandler<TEvent> : ICommandHandler<QueueCommand<TEvent>> where TEvent : IEvent
{
    public async Task ExecuteAsync(QueueCommand<TEvent> command, CancellationToken ct)
    {
        // handle command
        await Task.Delay(1000);
    }
}

// DI registration
var app = bld.Build();
app.Services.RegisterGenericCommand(typeof(QueueCommand<OrderCreatedEvent>), typeof(QueueCommandHandler<OrderCreatedEvent>));
// or with the added overload
app.Services.RegisterGenericCommand<QueueCommand<OrderCreatedEvent>, QueueCommandHandler<OrderCreatedEvent>>();

// Usage
public override async Task HandleAsync(ExtraRequest req, CancellationToken ct)
{
    var queueCommand = new QueueCommand<OrderCreatedEvent>()
    { 
        Event = new OrderCreatedEvent() 
    };

    await queueCommand.QueueJobAsync();

    await SendAsync(req);
}
```

### DIFF
--- a/Src/Library/Messaging/Commands/CommandExtensions.cs
+++ b/Src/Library/Messaging/Commands/CommandExtensions.cs
@@ -1,4 +1,4 @@
-﻿using System.Collections.Concurrent;
+using System.Collections.Concurrent;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace FastEndpoints;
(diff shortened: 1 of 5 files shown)


dj-nitehawk:
looks alright at first glance.
add some test coverage and i think we can merge.
thanks!

dj-nitehawk:
good job and thank you very much 🙏

PR #899, opened 2025-02-27
Added CancellationToken overloads to Remote*Async calls

SjorsGoering:
Added CancellationToken overloads to RemoteExecuteAsync and RemotePublishAsync for convenience reasons. Without them users would need to pass

```cs
new CallOptions(cancellationToken: ct)
```
in order to pass along the CancellationToken.


dj-nitehawk:
did a few other changes as well. pls have a look and let me know if you see any issues with those.

btw, i'm thinking about parallelizing event handler executions for subscribers here:

https://github.com/FastEndpoints/FastEndpoints/blob/8b2a7311691ea1900207ff6d0f079021adaf0270/Src/Messaging/Messaging.Remote.Core/Client/Events/EventSubscriber.cs#L156

not sure if it's a good idea though. basically make it just like how the joq queue executes command handlers with a `Parallel.ForEachAsync()`. let me know if you have any thoughts regarding that.

thanks!

PR #911, opened 2025-03-10
drop NET 6 and NET 7 support

sciocoder:
drop NET 6 and NET 7 support and restore 'collection expressions | type alias' in generators


dj-nitehawk:
thank you so much for this. saved me some hours.
good job mate!

PR #912, opened 2025-03-14
Added extra regex source generators

SjorsGoering:
Added a regex source generator in the BaseEndpoint and 2 in the OperationProcessor for Swagger.
Noticed that v6 will only support net8.0+ so I updated also the nuget package description.

### DIFF
--- a/Src/Generator/Extensions.cs
+++ b/Src/Generator/Extensions.cs
@@ -1,33 +1,41 @@
-﻿#pragma warning disable IDE1006
-    using System.Text;
-    using System.Text.RegularExpressions;
-    using Microsoft.CodeAnalysis;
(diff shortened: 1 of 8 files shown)


dj-nitehawk:
can you rerun the benchmark with the memory diagnoser enabled? wanna see if the stringbuilder and the foreach causes more memory allocations than the compiled & cached regex.

btw, is you benchmark using a compiled & cached regex?

dj-nitehawk:
so, as i suspected, the cached/compiled regex is more memory efficient. i think the performance impact of regex is negligibe since it's still in the nanosecond range. even the char array solution is going to allocate close to twice that of regex. let's stick to regex where we can't use source genned regex yeah?

PR #918, opened 2025-03-21
Added TriggerJobExecution<TCommand> overload

SjorsGoering:
Added an an extra overload, with just the command type as parameter, to trigger job executions.  With this overload you don't need to keep a reference to an job instance just to trigger them all, which is convenient when queueing multiple jobs. See the change in the JobQueueTests unit test class how it would benefit.
What are your thoughts @dj-nitehawk?


dj-nitehawk:
looks good.
will merge in a couple of days when i get back.
thanks!

PR #920, opened 2025-03-24
Made Endpoint Verbs method with Http methods public available

SjorsGoering:
Made the Endpoint Verbs method with Http[] parameters public available and removed the one with string[] parameters, since in my opinion people should use the defined Http enum instead of passing in a magic string. Since the next version will be a major version bump 6.x I think a small breaking change is acceptable.
Maybe for completeness we could extend the Http enum with Connect and Trace since those are valid values for a Http Method, although I doubt FE users are using those. Let me know your thoughts about it @dj-nitehawk

### DIFF
--- a/Src/Library/Endpoint/Endpoint.Setup.cs
+++ b/Src/Library/Endpoint/Endpoint.Setup.cs
@@ -1,4 +1,4 @@
-﻿using System.Diagnostics.CodeAnalysis;
+using System.Diagnostics.CodeAnalysis;
 using System.Linq.Expressions;
 using System.Reflection;
 using System.Text.Json;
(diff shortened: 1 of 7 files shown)


Src/Library/Endpoint/Endpoint.Setup.cs
```diff
@@ -598,13 +590,7 @@ protected void Validator<TValidator>() where TValidator : IValidator
     /// <summary>
     /// specify one or more http method verbs this endpoint should be accepting requests for
     /// </summary>
-    protected void Verbs(params Http[] methods)
-        => Verbs(methods.Select(m => m.ToString()).ToArray());
-
-    /// <summary>
-    /// specify one or more http method verbs this endpoint should be accepting requests for
-    /// </summary>
-    public sealed override void Verbs(params string[] methods)
```
dj-nitehawk, line 607: we can't remove this method and constrain to the `Http` enum. This has to stay a `string[]` because there were some edge cases sometime back where people wanted to use http methods that we don't include in the enum. also if i remember correctly, some wanted to do custom verbs as well. so i don't think we can make this change.

dj-nitehawk:
yeah so i guess, we'll leave it as is?
i'll close this PR?

PR #921, opened 2025-03-26
Optimized item removal from collections

SjorsGoering:
Changed the logic in various places for removing an item from a collection from using an foreach with an extra .ToArray() (needed because collections can't be modified while iterating), to just a reversed for loop. This reduces the need for the extra .ToArray() call. 

Also switched from using the .Remove(...) method to .RemoveAt(...) because .Remove(...) internally first has to do determine the index of the item using an .IndexOf call. Since we have the index available we can just use it directly with the .RemoveAt(...) call.

@dj-nitehawk let me know your thoughts.


dj-nitehawk:
looks good... thank you very much 🙏

Issue #924, opened 2025-03-28
Add support for annotation attributes ProducesResponseType and SwaggerResponse


omon77:
As we discussed in the Discord, when an endpoint is annotated with a `Microsoft.AspNetCore.Mvc.ProducesResponseType` or `SwaggerResponse` attributes, they are being ignored and actually NSwag throws an error that the annotated status code is not in dictionary.

Discovered when trying to add 201/Created status code via annotation.

Proposing to add support, so that the annotation is treated sort of first-class citizen.

ProducesResponseType:
```csharp 
[HttpPost("/api/post201")]
[Microsoft.AspNetCore.Mvc.ProducesResponseType(StatusCodes.Status201Created)]
public sealed class CreatedEndpoint : EndpointWithoutRequest
{
    public override async Task HandleAsync(CancellationToken c)
    {
        await SendAsync("Created", StatusCodes.Status201Created);
    }
}
```

and SwaggerResponse

```csharp
[HttpPost("/api/post201")]
[SwaggerResponse(201)]
public sealed class CreatedEndpoint : EndpointWithoutRequest
{
    public override async Task HandleAsync(CancellationToken c)
    {
        await SendAsync("Created", StatusCodes.Status201Created);
    }
}
```

dj-nitehawk:
works out of the box now starting with `v6.0.0-beta.6`

PR #925, opened 2025-03-28
Add support for annotation attributes ProducesResponseType and SwaggerResponse #924

omon77:
Add support for annotation attributes `ProducesResponseType` and `SwaggerResponse` #924

SwaggerResponse is supported implicitly, as it generates ProducesResponseType metadata.

P.S. When an endpoint is annotated with 201, it will _NOT_ generate 204 for that endpoint.


dj-nitehawk:
went a slightly different direction.
starting with `v6.0.0-beta.6`, `ProducesResponseTypeAttribute` will now work out of the box.
also the automatic addition of responses will not happen now if the user has already added one.
so your initial requirement now works without an issue.

PR #938, opened 2025-04-22
Add generic interfaces for IRequestMapper<,> and IResponseMapper<,>

niekschoemaker:
Mapper now implements IResponseMapper<,> and IRequestMapper<,>

This allows code which wants to use the mappers to simply pattern match on the generic interfaces and then use them, instead of having to pattern match on Mapper, and then on Response/Request Mapper since the dependency trees for the mapper types are split, this resolves that and makes it one singular dependency tree.

This however might be considered an opiniated change, and I'm not sure if there's any implications for breaking changes for people relying on the specific interfaces, I did make sure that all interfaces still implement the simple marker interfaces, so nothing should break in that way.

But if this is not wanted in any way, feel free to simply close the pull request, I however do believe that this might be considered a better dependency layout for using the mappers.

### DIFF
--- a/Src/Library/Endpoint/Mapper/IRequestMapper.cs
+++ b/Src/Library/Endpoint/Mapper/IRequestMapper.cs
@@ -1,37 +1,45 @@
-﻿namespace FastEndpoints;
+﻿// ReSharper disable UnusedParameter.Global
+
+namespace FastEndpoints;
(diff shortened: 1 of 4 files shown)


dj-nitehawk:
looks good...
thank you very much!

PR #948, opened 2025-05-15
Add support for JsonSchemaTypeAttribute on request models

rudiv:
This PR adds support for adding `[JsonSchemaType(typeof(..))]` to your request models.

This solves a couple of cases, the particular one that created this PR is that it's impossible to define "string" as a constraint on route params, so for example:

```csharp
public class ReqDto {
   [BindFrom("id")]
   public long Id { get; set; }
}

Post("/something/{id:string}");
```

doesn't work.

This now does:

```csharp
public class ReqDto {
   [BindFrom("id"), JsonSchemaType(typeof(string))]
   public long Id { get; set; }
}

Post("/something/{id}");
```

(Note reason for the mismatch is, obviously, Javascript)


dj-nitehawk:
good one...
thank you very much 🙏

PR #950, opened 2025-05-23
Fix global generic post processor registration

vivecuervo7:
This fixes an issue with the global generic post-processor registration outlined [here](https://fast-endpoints.com/docs/pre-post-processors#global-processors).

Unable to create an issue due to vacation notice.


dj-nitehawk:
> fixes an issue with the global generic post-processor registration

what exactly is the issue you're having?
can you explain it so i can reproduce it please?

thanks!

vivecuervo7:
Sorry - I should have included a proper example!

Nice library, by the way. Been a pleasure to use - I'm going to try and convince the next team of devs I work with to run with this 😀

PR #971, opened 2025-07-13
Test helper methods serialize record query params as json

mbcrawfo:
Updates the test of query param helper methods to echo back and assert the actual deserialized value to ensure that records are handled correctly.


dj-nitehawk:
thank you!
good catch with the record tostring issue.
i've made a few minor tweaks and merged it.

PR #974, opened 2025-07-17
Added Func<T, string> overload to SendEventStreamAsync to differ the event name per item

SjorsGoering:
I have a use case where I want to stream back potentially different types of events back to a client using Server Sent Events. A client should call a SSE endpoint and be able to process various types of events when they arrive. Based on the event name it can determine how to deserialize the event data. However currently in FE there's no way to specify an eventName per event type, meaning that all events that are streamed back will have the same event type. With the current version I would need to either:
- Make my own SendEventStreamAsync implementation, or
- Let the client connect to multiple SSE endpoints (1 per type)
- Dynamically deserialize the event data to determine its type

Making it flexible in FE is, I would think also useful for other people and relatively easy to implement. So I made a PR that supports this use case.

@dj-nitehawk what are your thoughts?

### DIFF
--- a/Benchmark/FastEndpointsBench/CodeGenEndpoint.cs
+++ b/Benchmark/FastEndpointsBench/CodeGenEndpoint.cs
@@ -59,7 +59,7 @@ public class CodeGenEndpoint : Endpoint<CodeGenRequest, CodeGenResponse>
 
         //Logger.LogInformation("request received!");
         //validator is automatically being run by FastEndpoints
-        => SendAsync(
+        => Send.ResponseAsync(
             new()
             {
                 Id = req.Id,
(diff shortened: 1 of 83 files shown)


dj-nitehawk:
i think it would be cleaner if we introduced a wrapper record type such as this:
```cs
public record struct StreamItem(string EventName, object Item);
```
user can create the async enumerable source data like this:
```cs
private async IAsyncEnumerable<StreamItem> GetDataStream([EnumeratorCancellation] CancellationToken ct)
{
    while (!ct.IsCancellationRequested)
    {
        await Task.Delay(1000);

        var sec = DateTime.Now.Second;

        if (sec % 2 == 1)
        {
            yield return new("odd-second", Guid.NewGuid());
        }
        else
            yield return new("even-second", "some string...");
    }
}
```
that way all we need is an overload constrained to the `StreamItem` type, something like this:
```cs
SendEventStreamAsync(IAsyncEnumerable<StreamItem> eventStream, CancellationToken cancellation = default)
```

if you think that's a good idea, go ahead and make the changes...

thanks!

dj-nitehawk:
looks good...

`GetDataString` method would allow the user to customize the serialization in case they wanna use some other serializer. so i think it's good.

any chance you could add an integration test? no worries if you don't have the time.

PR #981, opened 2025-07-28
Made id and event in SSE optional and added extra StreamItem overload

SjorsGoering:
I have an use case where I want to stream back items (of various types) to the client. Some types that I want to return don't have the concept of an id. Of course I can just 'fix' it by doing:
```cs
yield return new StreamItem("42", notificationType, notification);
```

But checking the specs at: https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation I noticed that the id field is actually optional. So I've made some changes to the Server Sent Events implementation to make it a bit more flexible, to better support the standard and to prevent issues with reverse proxies.

- The id field is optional according to the standard, so I made it nullable in the StreamItem class
- I made the data field not-nullable because when omitting the data field or sending an empty string as value, the onmessage handler in the browser is not triggered.
- Added an optional Retry property
- Added the X-Accel-Buffering header which configures a reverse proxy such as NGINX to send data as its received rather than waiting for the entire (potentially endless) response.

What are your thoughts @dj-nitehawk ?

### DIFF
--- a/Src/Directory.Build.props
+++ b/Src/Directory.Build.props
@@ -1,7 +1,7 @@
 <Project>
     <PropertyGroup>
 
-        <Version>7.1.0-beta.2</Version>
+        <Version>7.1.0-beta.3</Version>
 
         <TargetFrameworks>net8.0;net9.0;net10.0</TargetFrameworks>
         <LangVersion>latest</LangVersion>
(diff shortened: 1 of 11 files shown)


dj-nitehawk:
looks good...
merged...
thank you!

PR #1008, opened 2025-10-21
Add support for binding request values from cookies

maxs-rose:
Added a new attribute `FromCookie` which will bind a model parameter to a cookie.

I have ended up working with a few odd API's that add their parameters into cookies. Whilst its currently possible to get the cookie in FE through the `HttpContext.Request.Cookies` its a bit tedious and also annoying to document through an OpenAPI spec.


maxs-rose:
I didn't update the tests just in case we you don't want this in FE. If you are happy with the setup will go ahead an finish off the rest of the tests :)

PR #1024, opened 2025-12-05
Add Jobs support to the FastEndpoints.Messaging package

SjorsGoering:
I saw the changes in FastEndpoints for creating a Messaging package, with commands and events, that can be used separately. The Messaging package however does not support JobQueues. A lot of solutions are structured in a way with multiple projects separating hosting from business logic. With this PR it is no longer needed to add the main FastEndpoints package (containing web/hosting dependencies) to business logic projects. Also it allows other projects such as WPF, Windows Forms or even Blazor WASM to use the JobQueues functionality.
Since developers might be adding the Messaging package to an Blazor WASM project I wanted to limit the addition of extra dependencies to keep the download size to a minimum, I solved this with 2 replacements:
- MemoryCache is replaced with an ConcurrentDictionary<Guid, CancellationTokenSource>
- The ApplicationStopping CancellationToken from the IHostApplicationLifetime is retrieved when initializing the JobQueue instance using reflection. If the type can't be resolved, the code is probably running in a Blazor WASM environment where there is no IHostApplicationLifetime support.

While making the changes I noticed that the CancelJobAsync method has a bug as far as I can see. The code inside the for loop expects the cts variable to not be null within ~10 seconds, however that seems impossible since cts is never reassigned inside the loop. Also the existing code might never remove null values added to the MemoryCache, because the SlidingExpiration is set to an cancelExpiry which is set by default to Timeout.InfiniteTimeSpan while initializing the JobQueue.

Let me know your thoughts @dj-nitehawk and @eeaquino

### DIFF
--- a/Src/Messaging/Messaging/Jobs/JobQueue.cs
+++ b/Src/Messaging/Messaging/Jobs/JobQueue.cs
@@ -109,14 +109,13 @@ sealed class JobQueue<TCommand, TResult, TStorageRecord, TStorageProvider> : Job
     //public due to: https://github.com/FastEndpoints/FastEndpoints/issues/468
     public static readonly string QueueID = _tCommandName.ToHash();
 
-    readonly ConcurrentDictionary<Guid, CancellationTokenSource?> _cancellations = new();
+    readonly ConcurrentDictionary<Guid, CancellationTokenSource> _cancellations = new();
     readonly ParallelOptions _parallelOptions = new() { MaxDegreeOfParallelism = Environment.ProcessorCount };
     readonly TStorageProvider _storage;
     readonly IJobResultProvider? _resultStorage;
     readonly SemaphoreSlim _sem = new(0);
     readonly ILogger _log;
     TimeSpan _executionTimeLimit;
-    TimeSpan _cancellationExpiry;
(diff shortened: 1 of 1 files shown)


dj-nitehawk:
i haven't looked at your code yet, but i think we shouldn't add the job queues to the "messaging" library. imho, job queues is not really a "messaging" feature. only commonality is the use of the `ICommand/Handler` interfaces from the `FastEndpoints.Messaging.Core` pkg. so, i believe job queues needs its own independant library called `FastEndpoints.JobQueues`, and it doesn't need to supprt WASM at all. can't imagine a UI app wanting to use job scheduling tbh. that way, we can keep using `MemoryCache` and it's auto expiry functionality, as well as `IHostApplicationLifetime` cancellation token and whatever else without much fuss. (i.e. leave `MemoryCache` as is pls.)

regarding the possible bug in `CancelJobAsync`, could we please do a separate PR for that one? when too many things are lumped into one PR it makes it a bit difficult to review.

SjorsGoering:
Indeed I agree that a separate package would be better. It also keeps things simpler, without needing to resort to reflection...
I'll publish a new PR when I have it ready.

PR #1025, opened 2025-12-05
Improve cleanup of obsoleted job cancellations

SjorsGoering:
The code inside the CancelJobAsync has a for loop that expects the cts variable to not be null within ~10 seconds, however that seems impossible since cts is never reassigned inside the loop. I think the intention of the loop was to retry retrieving the cts in order to cancel it.
Now that with this change no null values are added in the MemoryCache, there doesn't seem to be a reason any more to use a MemoryCache for its auto expiry functionality. Any added item in the ExecuteCommand method is removed in the end of the try block as well as in the catch. Also, the auto expiry functionality would effectively only work if someone would override the default Timeout.InfiniteTimeSpan value.
With this change converting the MemoryCache to a ConcurrentDictionary<Guid, CancellationTokenSource> makes more sense in my opinion.
What are your thoughts @dj-nitehawk

### DIFF
--- a/Src/Library/Messaging/Jobs/JobQueue.cs
+++ b/Src/Library/Messaging/Jobs/JobQueue.cs
@@ -118,7 +118,7 @@ sealed class JobQueue<TCommand, TResult, TStorageRecord, TStorageProvider> : Job
     readonly ILogger _log;
     TimeSpan _executionTimeLimit;
     TimeSpan _semWaitLimit;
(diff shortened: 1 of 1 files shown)


SjorsGoering:
I've restructured the datetime logic a bit in a way that the now + 5 minutes calculation is only executed once every 5 minutes, instead of every time the loop is executed.
I think it looks ok now and you can merge.

PR #1032, opened 2026-01-08
Fix:  FastEndpoints.Testing HttpClientExtensions.GetTestUrlFor

YoussefElshemi:
If the request model contains a null query parameter, it is still appended to the url as ?QueryParameter=, when really it should be omitted from the request url.

### DIFF
--- a/Tests/UnitTests/FastEndpoints/HttpClientExtensionsTests.cs
+++ b/Tests/UnitTests/FastEndpoints/HttpClientExtensionsTests.cs
@@ -22,11 +22,12 @@ public class HttpClientExtensionsTests
         var http = mockHttp.ToHttpClient();
         http.BaseAddress = new("http://localhost");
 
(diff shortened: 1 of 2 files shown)


Tests/UnitTests/FastEndpoints/HttpClientExtensionsTests.cs
```diff
@@ -42,6 +42,72 @@ await http.GETAsync<Endpoint, Request, Response>(
...
+            FromHeader = "fromHeader",
+            FromCookie = "fromCookie",
+            HasPermission = true
+        };
+
+        // Act
+        var testUrl = HttpClientExtensions.GetTestUrlFor<Endpoint, Request>(req, http);
+
+        // Assert
+        testUrl.ShouldBe("api/test/1/00000000-0000-0000-0000-000000000000/stringValue/{fromClaim}/{fromHeader}/{hasPermission}");
```
YoussefElshemi: previously this would've been
api/test/1/00000000-0000-0000-0000-000000000000/stringValue/{fromClaim}/{fromHeader}/{hasPermission}?fromQuery=

PR #1040, opened 2026-01-13
Reduce GetNextBatch calls in JobQueue

SjorsGoering:
I noticed that the CommandExecutorTask method in the JobQueue class tries to fetch the next batch of jobs more often than needed in 2 scenarios.

When the _sem semaphore is released once, the CommandExecutorTask continues after the _sem.WaitAsync call, checking if _cancellations need to be cleaned and then restarting the loop which tries to fetch the next batch of jobs. If there are records they are executed in parallel and the _cancellations are cleaned if needed and the loop is restarted. Basically for every _sem.Release the GetNextBatch method is called twice.

Another scenario is when multiple threads are more or less simultaneously releasing the _sem semaphore. Then for every Release call the _sem.WaitAsync continuous instantly, effectively causing the GetNextBatch to be called twice per release call. But if the jobs were triggered at the same moment there's a high change that the first GetNextBatch call already returned all of them, making successive calls unneeded.

The solution that I'm proposing is to
- First check if there are records to process, if so execute them 
- Only if the number of records is less then the batch size let the current thread wait
- If the number of records equals or more than the batch size then continue the loop to fetch the next batch
- Reset the _sem CurrentCount to 0 to ignore multiple concurrent Releases

Instead of using a SemaphoreSlim, I thought about using a ManualResetEventSlim. However that implementation has no async Wait support.

Let me know what you think @dj-nitehawk

### DIFF
--- a/Src/JobQueues/JobQueue.cs
+++ b/Src/JobQueues/JobQueue.cs
@@ -288,7 +288,7 @@ sealed class JobQueue<TCommand, TResult, TStorageRecord, TStorageProvider> : Job
             var recordsCount = records.Count();
             if (recordsCount > 0)
                 await Parallel.ForEachAsync(records, _parallelOptions, ExecuteCommand);
-            
+
             if (recordsCount < _parallelOptions.MaxDegreeOfParallelism)
             {
                 // less records than page size, so wait on the semaphore before next iteration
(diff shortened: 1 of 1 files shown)


dj-nitehawk:
good job!

PR #1046, opened 2026-01-15
Fix: GetTestUrl - Correct Date/Time strings

YoussefElshemi:
If the query parameter is a date/time object, when to string is called, it uses local culture which does not serialize correctly on the request side, so this is to keep it in the correct format

### DIFF
--- a/Src/Library/Testing/HttpClientExtensions.cs
+++ b/Src/Library/Testing/HttpClientExtensions.cs
@@ -1,6 +1,7 @@
 // ReSharper disable InconsistentNaming
 
 using System.Collections;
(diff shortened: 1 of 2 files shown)


YoussefElshemi:
> can you try my change?
> 
> [1ebbb4d](https://github.com/FastEndpoints/FastEndpoints/commit/1ebbb4d53f9649ab72e0a7954fb54ae5474b448e)
> 
> does it solve the issue?

hi, no that doesnt fix it, i realised in my change the issue isnt the culture, its just needs a specific format 'o' to be iso format

PR #1058, opened 2026-01-31
NativeAOT : PR1: Core AOT Infrastructure for Generic Processors

Pinox:
@dj-nitehawk  Not sure if useful, use as you please.

### Problem
Open generic pre/post processors fail in Native AOT mode because `MakeGenericType()` + `Activator.CreateInstance()` don't work without proper annotations or code generation.

### Solution
- **GenericTypeRegistryProvider**: Factory lookup methods for closed generic processors
- **GenericTypeRegistryGenerator**: Source generator that discovers open generic processors and generates factory dictionaries at compile-time
- **EndpointDefinition**: Uses AOT-safe factory lookups with fallback to reflection

### Files Changed (5)
1. `NativeAotChecker/Endpoints/GenericProcessorEndpoint.cs` - Test endpoint
2. `Src/Core/GenericTypeRegistry/GenericTypeRegistryProvider.cs` - Factory registry
3. `Src/Generator/GenericTypeRegistryGenerator.cs` - Source generator
4. `Src/Library/Endpoint/Auxiliary/EndpointDefinition.cs` - Integration
5. `Tests/NativeAotTests/NativeAotCheckerTests/EndpointTests.cs` - Test

### Testing
- Without this PR: App crashes with `InvalidOperationException` 
- With this PR: Generic processors work correctly in AOT mode

### DIFF
--- a/NativeAotChecker/Endpoints/GenericProcessorEndpoint.cs
+++ b/NativeAotChecker/Endpoints/GenericProcessorEndpoint.cs
@@ -1,0 +1,110 @@
+using System.Text.Json.Serialization;
+
+namespace NativeAotChecker.Endpoints;
+
+// ============================================================================
+// GENERIC PROCESSOR AOT TEST
+// ============================================================================
+// This test validates that OPEN GENERIC pre/post processors work in AOT mode.
+// 
+// WITHOUT the source generator (PR1):
+//   - The runtime uses Activator.CreateInstance to instantiate 
+//     AotGenericPreProcessor<GenericProcessorRequest> from the open generic type
+//   - In AOT mode, this FAILS because MakeGenericType + CreateInstance is not supported
+//   - The endpoint will NOT run the processor and fail silently or throw
+//
+// WITH the source generator (PR1):
+//   - The generator creates factories like:
+//     { (typeof(AotGenericPreProcessor<>), typeof(GenericProcessorRequest)), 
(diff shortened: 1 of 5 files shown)


dj-nitehawk:
it's just that i'm going through the warnings with surgical precision to avoid unnecessarily complicating/messing up the code base. as a result, i only had to use `[DynamicallyAccessedMembers(...)]` in just one place. 99% of the warnings turned out to be false positives, or was easily fixed by adding another interface to the whitelist in the "discovered types generator".

so... it would be great if you could simply repro any actual features that aren;t working in aot by adding a failing test case to the "checker" app. i will then dig in and find a solution without cluttering the source.

sounds like a plan?

Pinox:
@dj-nitehawk 

Is something like this not acceptable for only AOT scenario's ?  This should not effect any JIT library users, only those users that would prefer to run everything source generated. I would suspect JIT users would be the primary beneficiaries here as their FE implementation would be even faster than they already are.  

```csharp
.AddFastEndpoints(o =>
{
    // Type discovery (already exists)
    o.SourceGeneratorDiscoveredTypes = DiscoveredTypes.All;
    
    // Hypothetical additional AOT flags:
    o.UseSourceGeneratedReflection = true;  // Auto-call ReflectionCache.AddFrom*()
    o.UseSourceGeneratedValidators = true;  // Source-gen validator registration
    o.UseSourceGeneratedBinders = true;     // Source-gen custom binders
})
```

PR #1065, opened 2026-02-05
test(aot): Add failing test for generic command handlers

Pinox:
This test demonstrates that generic command handlers fail in Native AOT mode.

Root Cause:
- Generic command handlers require runtime type construction
- MakeGenericType() calls fail in AOT
- Service provider cannot resolve open generic types at runtime

Test Endpoint: GenericCommandHandlerEndpoint.cs
Test: GenericCommandHandlerTest.cs


dj-nitehawk:
migrating to linux took a whole lot more time than expected. will be looking in to this today.

Pinox:
which linux os did you go for ? Fedora or Pop!_OS or another one

Pinox:
zorin & kde plasma probably the most windows like for linux but the linux ui's are becoming better every day. Personally not in a rush to switch as I like Windows UI visually but technically and with rise of ai I think the linux distro's are going to become an easy switch going forward.

This is a custom kde windows lookalike and it looks damm good if you ask me.

image

PR #1086, opened 2026-04-09
Fix package versioning and remove weird reflection

dhypa:
I saw this article: https://devblogs.microsoft.com/dotnet/aspnet-core-2-3-end-of-support/
I try to help in removing the old dependency. Hopefully this will stop old dependency showing up on nuget website for the dotnet 10 version of FastEndpoints.Core:
image

I also remove reflection from your newtonsoft -> STJ option converter. I couldn't manage to do it for one of the functions because there is no public API on the class that I could access.


dj-nitehawk:
thanks for the PR.
i dropped support for `netstandard2.1` and made a few more cleanups.
i'll merge this now.
if you find any issues, please create a new PR.
thank you!

PR #1087, opened 2026-04-16
Fixed FastEndpoints.Generator not seeing types defined in FastEndpoints.Attributes

SjorsGoering:
When using the FastEndpoints.Generator in a (business layer) project that only has a reference to FastEndpoints.Messaging.Core I get compile errors in the GeneratedReflection about the ReflectionCache that cannot be found. I think this is because the Source Generator cannot see types defined in referenced projects.
https://github.com/dotnet/roslyn/issues/57997

I took some inspiration in making a fix from here:
https://andrewlock.net/creating-a-source-generator-part-8-solving-the-source-generator-marker-attribute-problem-part2/

Ps. Not related, but I found that the AccessControlGenerator assumes that people have global usings defined for System, System.Collection.Generic and Linq. I therefore added them explicitly so people don't need to worry about that.


dj-nitehawk:
will review soon. thanks!

did you try referencing `FastEndpoints.Attributes` in your business project along with `FastEndpoints.Attributes` by any chance?

if that works, maybe that's a better solution than packing the dlls in the generator pkg?

SjorsGoering:
Referencing FastEndpoints.Attributes in the business project along FastEndpoints.Generator works. However that this is needed is not very obvious (and at the moment requires people to check the FE source code). Instead of documenting this requirement somewhere I think it's easier to just fix the nupkg so that the FastEndpoints.Attributes is also available by Roslyn. Since the fix is pretty simple (copying the dll also to the analyzers/dotnet/cs folder) I would say going that route makes more sense.

dj-nitehawk:
alright then...
will merge soon.
thank you!

dj-nitehawk:
btw, if you have some free time, i could really use another set of eyes on the new `FastEndpoints.OpenApi` package.

pls see [the changelog](https://github.com/FastEndpoints/FastEndpoints/blob/main/Src/Library/changelog.md) for the details and the [preview docs](https://dev.fastendpoints-doc-site.pages.dev/docs/openapi-documents) on how to use.

would be good if you can switch out `FastEndpoints.Swagger` in one of your largest projects and see if the [new lib](https://www.nuget.org/packages/FastEndpoints.OpenApi/8.2.0-beta.11) generates correct openapi.json content.

thanks!

SjorsGoering:
Sure, I'll have a look at the docs and the code and give it a try with a project.

dj-nitehawk:
copying to the analyzer folder is most likely unnecessary.
could you give `v8.2.0-beta.20` a try pls.

SjorsGoering:
I checked with v8.2.0-beta.20 and as you suspected it does not fix the issue. I will check later how to solve it.

dj-nitehawk:
can you try `v8.2.0-beta.21`
i added back the copy to analyzer folder.

SjorsGoering:
I tested it and that did not work either. In the meantime I've managed to come up with some changes that locally work for me, see:
https://github.com/FastEndpoints/FastEndpoints/pull/1091

PR #1090, opened 2026-04-22
Fixed handling empty strings and trailing/leading slashes in route prefixes

SjorsGoering:
I have an API that retrieves the route prefix from an IConfiguration, something like this:
```cs
app.UseFastEndpoints(c =>
{
    c.Endpoints.RoutePrefix = app.Configuration.GetValue<string?>("RoutePrefix");
});
```

When specifying null as the RoutePrefix value and calling GetValue<string?>("RoutePrefix") I get "" as result (as opposed to null). According to https://learn.microsoft.com/en-us/dotnet/core/compatibility/extensions/10.0/configuration-null-values-preserved in .net 10 this behavior is changed so that it does return null.

Since I'm not running on .net 10 (yet) I get the following exception when running the app:
"The route template separator character '/' cannot appear consecutively. It must be separated by either a parameter or a literal value."

I'm able to workaround this in my codebase by swapping a "" value with null. However I think it would be better to make the BuildRoute method of FastEndpoints more forgiven, which I did in this PR.


dj-nitehawk:
improved and merged.
thank you!

PR #1091, opened 2026-04-23
Fixed FastEndpoints.Generator not seeing types referenced in FastEndpoints.Attributes

SjorsGoering:
Attempt 2 in trying to fix the FastEndpoints.Attribute reference issues when using FastEndpoints.Generator. 
I've created a nuget package locally and tested it in my project and it compiles successfully.

Setting DevelopmentDependency to false allows Roslyn to see any referenced projects. It made my project compile again. Setting it to false makes sense since the Generator package is not only a development dependency, since the Attributes dll is supposed to end up in the output bin folder.

However that will also put the FastEndpoints.Generator dll in the output bin folder, which is not needed. To prevent that I've changed the BuildOutputTargetFolder to the analyzers\dotnet\cs folder, making a bunch of None Include="x" statements unnecessary.

What's left is to copy the FastEndpoints.Attributes.dll into the lib\netstandard2.0 folder so that dll does end up in the output bin folder.


dj-nitehawk:
i think this is the solution. merging now.

PR #1092, opened 2026-04-24
feat: expose FastEndpoints as MCP tools and A2A skills

EelcoLos:
Adds two addon packages that bridge FastEndpoints into agentic protocols, with **zero AI-specific public API surface in the main library**. Shared plumbing (endpoint invoker, JSON-Schema builder, FluentValidation enricher) is compile-included as `internal` into each addon rather than shipped as a third package.

* **`FastEndpoints.Mcp`** `1.0.0-beta.1` — Model Context Protocol tools over stdio + Streamable HTTP, built on the official `ModelContextProtocol` C# SDK.
* **`FastEndpoints.A2A`** `0.1.0-preview.1` — A2A skills with an Agent Card at `/.well-known/agent.json` and a JSON-RPC 2.0 dispatcher. Protocol still stabilising under the Linux Foundation; hence the preview tag.

Both addons are **versioned independently** of FastEndpoints core — the addon release cadence is not tied to the main library.

## Addressing the owner's constraint

On Discord: *"ideally i'd like any AI related stuff to be not integrated into the main library itself. (even if it means AI lovers will have to do `this.Whatever()` to configure it.) it's ok to have internal state/metadata related to AI stuff, but certainly no public api surface."*

This PR now complies:

* No `McpToolInfo` / `A2ASkillInfo` types, no `[McpTool]` / `[A2ASkill]` attributes, no `McpTool(...)` / `A2ASkill(...)` methods in `FastEndpoints` or `FastEndpoints.Attributes`. All four AI files were deleted and the three core files that had been modified were reverted.
* `FastEndpoints.xml` contains **zero** `Mcp|A2A|Agent` matches after build — the public doc surface is clean.
* The only concession kept in core is the `InternalsVisibleTo` entries for the two addon assemblies (in `Src/Library/Metadata.cs`). `EndpointInvoker` legitimately needs access to `BaseEndpoint.Definition` / `.HttpContext` setters and `ExecAsync(ct)` to run an endpoint through the real FE pipeline from outside an ASP.NET request — and the owner explicitly green-lit "internal state/metadata".

## Opt-in surface

Fluent form inside `Configure()`. If the endpoint already has an OpenAPI `Summary` set, calling `McpTool()` / `A2ASkill()` with no arguments is enough — the tool name and description are derived automatically:

```csharp
{
    Post("/products");
    Summary(s =>
    {
        s.Summary     = "Add Product";                           // -> MCP tool name "add_product"
        s.Description = "Adds a new Product to a Product Group"; // -> tool description
    });
    this.McpTool();    // FastEndpoints.Mcp -- no duplication needed
    this.A2ASkill();   // FastEndpoints.A2A
}
```

Name and description can always be overridden explicitly:

```csharp
this.McpTool("custom_name", "Custom description.");
this.A2ASkill("custom_id", tags: ["orders"]);
```

If neither an explicit name/description nor a `Summary` is set, a `LogWarning` fires at startup pointing to exactly what to add — fast-fail developer experience without breaking existing code.

The `this.` prefix is **required by the C# language** — `McpTool` / `A2ASkill` are extension methods on `BaseEndpoint`, and the C# compiler only searches for extension methods when there is an explicit receiver. Bare `McpTool(...)` inside `Configure()` does not resolve. The alternative was to ship parallel `McpEndpoint<,>` / `A2AEndpoint<,>` base classes, forcing an inheritance choice on every opt-in endpoint to save four characters; we keep the simpler extension form, which is the line the owner endorsed on Discord.

Attribute form (for attribute-configured endpoints):

```csharp
[McpTool("get_order", Description = "Retrieve an order by id.")]
public class GetOrder : Endpoint<GetOrderRequest, Order> { ... }
```

Authors composing configuration from helpers can also call `Definition.McpTool(...)` / `Definition.A2ASkill(...)` (extension on `EndpointDefinition`) directly.

Behind both forms, the addon stores a single `McpToolInfo` / `A2ASkillInfo` in the endpoint's existing public `EndpointDefinition.EndpointMetadata` bag (fluent) or reads the addon attribute back from `EndpointDefinition.EndpointAttributes` (attribute). Both are generic, AI-agnostic extension points FastEndpoints already exposes to every addon — no new hooks added to core.

## Wire-up

```csharp
builder.Services.AddFastEndpoints()
                .AddMcp()
                .AddA2A(o =>
                {
                    o.AgentName = "orders-agent";
                    o.Provider  = new() { Organization = "ECorp" };
                });

var app = builder.Build();

app.UseFastEndpoints()
   .UseMcp()                                              // /mcp
   .UseA2A();                                             // /a2a + /.well-known/agent.json

// with route conventions:
app.UseFastEndpoints()
   .UseMcp("/mcp", route => route.RequireAuthorization())
   .UseA2A(configureRpcRoute: r => r.RequireAuthorization("a2a-caller"));
```

## Package layout

```
Src/Agents/
  Directory.Build.props      shared PackageTags; imports parent Src/Directory.Build.props
  README.md                  addon design notes (in-repo, not shipped)
  Shared/                    compile-included into both addons as internal types
    EndpointInvoker.cs
    JsonSchemaBuilder.cs
    FluentValidationSchemaEnricher.cs
    NamingHelpers.cs           ToSnakeCase -- PascalCase + space/hyphen input (shared by Mcp + A2A)
    AssemblyAttributes.cs      InternalsVisibleTo Int.Agents for each addon
  Mcp/                       FastEndpoints.Mcp 1.0.0-beta.1
  A2A/                       FastEndpoints.A2A 0.1.0-preview.1
```

`Shared/*.cs` files are compile-included into both addon csprojs as internal types, so a consumer who installs both packages still sees only one copy of each type in their own code (one per addon assembly, both internal). No public type collisions.

## What's inside each addon

* JSON Schema generation via STJ `JsonSchemaExporter`, enriched with FluentValidation rules (`required`, `minLength`/`maxLength`, `pattern`, `minimum`/`maximum`, `format: email`).
* `EndpointInvoker` runs endpoints through the full FE pipeline via a scoped `DefaultHttpContext` with a `MemoryStream` body. Validation failures surface as structured errors agents/LLMs can self-correct from.
* MCP: tool list + call handlers, per-session `ToolVisibilityFilter` hook, configurable `ToolFilter`, optional `IncludeOutputSchemas`.
* A2A: agent card builder, `message/send` JSON-RPC handler.

## Scope boundaries (MVP)

* MCP stretch items deferred: resources/prompts exposure, streaming responses, end-to-end per-session auth visibility propagation through the SDK's per-session options.
* A2A shipped as `0.x-preview`. Sync `message/send` only. Streaming, task lifecycle (`tasks/get`/`tasks/cancel`), proof-of-intent, push-notification config, and federated discovery are roadmap items.

## Test plan

* [x] `dotnet build FastEndpoints.slnx` -- 0 errors.
* [x] `dotnet test Tests/IntegrationTests/FastEndpoints.Agents` -- 23/23 passing (schema builder, FluentValidation enricher, NamingHelpers).
* [x] `dotnet pack Src/Agents/Mcp/FastEndpoints.Mcp.csproj` produces `FastEndpoints.Mcp.1.0.0-beta.1.nupkg` independently of the core NuGet version.
* [x] `dotnet pack Src/Agents/A2A/FastEndpoints.A2A.csproj` produces `FastEndpoints.A2A.0.1.0-preview.1.nupkg` independently.
* [x] `grep "Mcp|A2A|Agent"` on `FastEndpoints.xml` -> zero matches.
* [ ] Follow-up PR: end-to-end MCP client round-trip via `ModelContextProtocol.Client` against an in-proc server.
* [ ] Follow-up PR: A2A `agent.json` + JSON-RPC round-trip integration test.
* [ ] Manual: connect Claude Desktop / VS Code to a sample app running `AddMcp()` and invoke a tool.

## Commits

Twenty small commits, each independently reviewable:

1.  `a5a32dc5` initial core opt-in metadata (later reverted in commit 9)
2.  `73d5b7e7` scaffold `FastEndpoints.Agents.Core` (later folded into Mcp + A2A in commit 8)
3.  `589ae69d` scaffold `FastEndpoints.Mcp`
4.  `a94ba256` scaffold `FastEndpoints.A2A` preview
5.  `9543cd9f` `InternalsVisibleTo` + MCP SDK API corrections
6.  `e1e65d20` unit tests for schema builder + FV enricher
7.  `91183331` chainable `AddMcp`/`AddA2A` + `UseMcp`/`UseA2A` naming
8.  `1de95fce` fold `Agents.Core` into Mcp + A2A as shared-compile internals (drops the third package)
9.  `ba5b81cd` remove AI-specific public surface from FastEndpoints core
10. `a3b62984` version addon packages independently of core (Mcp `1.0.0-beta.1`, A2A `0.1.0-preview.1`)
11. `15edb3fa` `Src/Agents/README.md`
12. `8bf82b93` docs: call out the `this.` prefix as a C# language requirement
13. `748e4b57` drop incidental whitespace diffs in `EndpointDefinition` / `Endpoint.Setup`
14. `c43da351` trim unused `ModelContextProtocol` / `Microsoft.Extensions.AI` CPM entries
15. `e653fa16` report validation failures regardless of response status code
16. `ec4c7ae8` openapi lib improvements
17. `c82e19f4` derive MCP tool name from OpenAPI Summary; move `ToSnakeCase` to shared `NamingHelpers`; startup `LogWarning` when no name/description resolvable
18. `f5593f0c` derive A2A skill id/name from OpenAPI Summary; fix missing `ToSnakeCase` on A2A type-name default; startup `LogWarning`
19. `3fe8d384` `NamingHelpersTests` (23 cases); fix CA2017 log-template warnings
20. `b2dc4e97` fix `A2ASkillDispatcher.FindSkill()` to use the same id fallback chain as `AgentCardBuilder` (explicit > Summary > type name), preventing runtime "skill not found" when no explicit id is set


EelcoLos:
we've discussed: going to seperate it from the core library. working on that

AndersFredborg:
Hi,

Thanks, this is a great initiative bringing MCP to FastEndpoints :)

I have 1 suggestion i would like to make. 

Often when we write a fast endpoint, we also specify the OpenAPI details of the endpoint using the Summary operator like this:

```
{
    Post(EndpointRoutes.Oee.Products.Add);
    Group<OeeGroup>();
    Validator<AddProductValidator>();
    Description(x => x
        .Produces<ProductResponse>(StatusCodes.Status200OK)
        .Produces<ProblemDetails>(StatusCodes.Status400BadRequest)
        .Produces<ProblemDetails>(StatusCodes.Status426UpgradeRequired));
    Summary(s =>
    {
        s.Summary = "Add Product";
        s.Description = "Adds a new Product to a Product Group";
        s.Responses[StatusCodes.Status426UpgradeRequired] = "The OEE feature is not enabled in the platform";
        s.Responses[StatusCodes.Status400BadRequest] = "Validation failed or failed to create the product";
    });
}
```

In the code you use the syntax: this.McpTool("get_order", "Retrieve an order by id.");  to specify the details for the tool.

```
{
    Get("/orders/{id}");
    this.McpTool("get_order", "Retrieve an order by id.");   // FastEndpoints.Mcp
    // or: this.A2ASkill("get_order", tags: ["orders"]);     // FastEndpoints.A2A
}
```

But if the Tool allready has a defined Summary for the OpenAPi Spec, we should be able to just re-use that?

Eg. use:
        s.Summary = "Add Product";
        s.Description = "Adds a new Product to a Product Group";

So instead it just becomes:
```
{
    Get("/orders/{id}");
    this.McpTool();   // FastEndpoints.Mcp
}
```
It then uses the Summary and Description unless overrided (lowercasing the summary and replaces spaces with '_')
McpTool Name = "add_product"
McpTool Description = "Adds a new Product to a Product Group"

That way we dont have to duplicate our documentation of the endpoint, and it gets easier to maintain with less code. However you could add support to override the Name and Description of the Tool.

Does this make sense?

Thanks.

EelcoLos:
Hi, thanks for the thoughtful suggestion. This is a great DX improvement!

You're right that there's no reason to duplicate the documentation. Here's what already works and what I'll add:

 **Already works today:** `Description` already falls back to `EndpointSummary.Description`, so the long description is free if you have `Summary(s => s.Description = ...)` set.

 **What I'll add:**
 - MCP `Name` will fall back to `EndpointSummary.Summary` converted to snake_case. So `"Add Product"` becomes `"add_product"`. The `ToSnakeCase` helper already handles PascalCase type names; I'll extend it to collapse spaces and hyphens too.
 - A2A `Name` (the display name shown in the agent card) will also fall back to `EndpointSummary.Summary` as-is.
 - A startup warning will be logged whenever a tool/skill falls all the way through to the type name, so developers get a clear nudge if they forgot to set a `Summary`.

 **One note on A2A `Id` specifically:** unlike the display name, `Id` is meant to be a stable, machine-readable identifier. It's how the dispatcher routes incoming skill calls. Deriving it automatically from the human-readable Summary string could cause subtle breakage if someone later renames `s.Summary` (the id changes, existing clients break). So for A2A I'll keep `Id` defaulting to the endpoint type name (snake_cased), and only the display `Name` and `Description` will pick up from Summary. You can always set an explicit id if you want a cleaner one: `this.A2ASkill("add_product")`.

 **Final result for MCP:**
 ```csharp
 Summary(s => {
     s.Summary = "Add Product";                            // → MCP name "add_product"
     s.Description = "Adds a new Product to a Product Group";  // → MCP + A2A description
 });
 this.McpTool();   // no duplication needed
 ```
 Override still fully supported: `this.McpTool("my_tool", "Custom description")` takes precedence over Summary.

 I implemented this on the branch and updated the PR body.

dj-nitehawk:
i think the current state is good enough for the first beta. will push it to nuget soon.
thanks for taking the initiative to get this going 🙏

Issue #1089, opened 2026-04-22
Malformed multipart request body throws unhandled IOException / InvalidDataException (500) instead of 400


neotrow:
First off — huge thanks for FastEndpoints. It's been a joy to work with and has completely replaced controllers for us. 

We ran an OWASP ZAP security scan against our API and discovered a behavior that produces false-positive SQL injection alerts. We dug into it and found what looks like a gap in how malformed `multipart/form-data` requests are handled during model binding. Happy to submit a PR if the fix direction below looks good to you!

## Summary

When a client sends a malformed or prematurely truncated `multipart/form-data` request body, `RequestBinder` throws an unhandled exception that escapes the endpoint pipeline as a 500 Internal Server Error. This is unexpected because FastEndpoints already handles other binding failures gracefully (returning 400 via ProblemDetails). Malformed multipart is a client error and should result in a 400, not a 500.

## Exception types observed

Two distinct exceptions are thrown from `RequestBinder.BindFormValues` depending on the nature of the malformed input:

1. **`System.IO.IOException`** — thrown when the request body is truncated mid-stream:
   ```
   TYPE: IOException
   REASON: Unexpected end of Stream, the content may have already been read by another component.
   ```
   Stack trace origin: `Microsoft.AspNetCore.WebUtilities.MultipartReaderStream.ReadAsync`

2. **`System.IO.InvalidDataException`** — thrown when a multipart section has a malformed `Content-Disposition` header:
   ```
   TYPE: InvalidDataException
   REASON: Form section has invalid Content-Disposition value:
   ```
   Stack trace origin: `Microsoft.AspNetCore.Http.Features.FormFeature.InnerReadFormAsync`

Both exceptions propagate through:
```
FastEndpoints.RequestBinder`1.BindFormValues
FastEndpoints.RequestBinder`1.BindAsync
FastEndpoints.Endpoint`2.BindRequestAsync
FastEndpoints.Endpoint`2.ExecAsync
```

## Expected behavior

`IOException` and `InvalidDataException` thrown by the ASP.NET Core form reader are semantically a binding failure caused by a bad client request. They should result in a 400 response via ProblemDetails, consistent with all other binding failures in `RequestBinder`.

[...]

PR #1093, opened 2026-04-27
Malformed multipart request body throws unhandled IOException / InvalidDataException (500) instead of 400

neotrow:
Fixes issue #1089 

## Why unconditional catch instead of a default `FormExceptionTransformer`?

I considered two approaches:

### Option A (chosen): Catch `IOException`/`InvalidDataException` unconditionally

When `FormExceptionTransformer` is `null` (the default), wrap `Execute()` in a try/catch for these two specific exception types and add them as validation failures.

### Option B (rejected): Provide a default `FormExceptionTransformer`

Give `FormExceptionTransformer` a non-null default (similar to how `JsonExceptionTransformer` already has one), so the existing catch-all block handles everything.

### Why I went with Option A

1. **Re-throw problem**: The existing `FormExceptionTransformer` catch block unconditionally calls `ctx.ValidationFailures.Add(transformer(e))` for *every* exception. If I gave it a default transformer, it would swallow unexpected exception types (e.g. `OutOfMemoryException`, `StackOverflowException`) as validation failures instead of letting them propagate. To avoid that, I'd need either:
   - A nullable return type (`Func<Exception, ValidationFailure?>`) — breaking API change
   - Re-throw logic inside the transformer using `ExceptionDispatchInfo` — awkward, resets stack trace concerns
   - A null-check after the transformer call in `BindFormValues` — changes shared code path behavior

2. **Surgical scope**: Option A only affects the `null` branch (the default path). The `else` branch — used when someone has configured `FormExceptionTransformer` — is completely untouched. Zero risk of regression for users who already use this feature.

3. **Specificity**: I only catch the two exception types that are unambiguously client errors (`IOException` from truncated bodies, `InvalidDataException` from malformed headers). Everything else still propagates as before.

4. **No API surface changes**: No new properties, no signature changes, no breaking changes.

---

## Question for maintainer: `ExcludeInCiCd` trait on tests

The two new tests in `MalformedMultipartTests.cs` are currently marked with `[Trait("ExcludeInCiCd", "Yes")]`, matching the pattern used by the existing `MaxRequestBodyLimitTests`. These tests spin up a standalone `WebApplication` on a hardcoded port (1025).

**Should these tests be included in CI/CD or excluded?**

- **Keep `ExcludeInCiCd = Yes`**: Consistent with `MaxRequestBodyLimitTests`. Avoids potential port conflicts in CI. Tests can be run manually.
- **Remove the trait**: Tests run in CI for continuous coverage. Port conflicts are unlikely but possible if tests run in parallel.

Let me know your preference and I'll adjust accordingly.


dj-nitehawk:
impoved and merged. thank you!

PR #1103, opened 2026-05-15
Disposing SigningKey after update

SjorsGoering:
With some help I found a memory leak in the JwtSigningOptions.

- JwtSigningOptions.UpdateSigningKey() previously replaced the static _securityKey field with a new RsaSecurityKey and abandoned the old one with a //do not dispose comment. On each key rotation the old RSA instance — which wraps an unmanaged OS crypto handle — was leaked.
- The new _securityKey is now swapped in via Interlocked.Exchange, and the displaced key's RSA handle is disposed immediately after. The first call is safe (previous value is null) and a style change from asymmetric to symmetric is also safe (the cast to RsaSecurityKey returns null).

I also extended the .gitignore for .lscache files which are generated by the C# Dev Kit extension in VSCode.


dj-nitehawk:
apparently, we don't need to hold on to a live `RSA` by doing `rsa.ExportParameters()` when creating a `RsaSecurityKey`.
have a look at my change and let me know what you think.

SjorsGoering:
Indeed, that seems true and makes the change simpler. It looks good to me and you an merge.

PR #1104, opened 2026-05-15
Aligned [HasPermission] binder to use case-sensitive permission value comparison

SjorsGoering:
With some help I found a casing check mismatch between auth policies and permission bindings.

BindHasPermissionProps() in RequestBinder compared permission claim values using OrdinalIgnoreCase, while the authorization policies built in MainExtensions.AddSecurityPolicy() use Ordinal (case-sensitive) for the same values — both AllowAnyPermission (line 338) and AllowAllPermissions (line 346). This meant the [HasPermission] DTO binder and the endpoint authorization gate made opposite assumptions: a token carrying "Admin" would be rejected by an auth policy requiring "admin", but Request.IsAdmin would still be set to true by the binder.
- Changed the value comparison in BindHasPermissionProps from OrdinalIgnoreCase to Ordinal to match the auth policy.

I also extended the .gitignore for .lscache files which are generated by the C# Dev Kit extension in VSCode.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ BenchmarkDotNet.Artifacts/
 *.tmp_proj
 *.log
 *.vspscc
+*.lscache
 publish/
 *.nupkg
 *.snupkg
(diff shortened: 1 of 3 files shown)


dj-nitehawk:
thank you for these PRs.

PR #1106, opened 2026-05-18
Add IStreamCommandHandler support with middleware pipelines

SjorsGoering:
Currently I'm using a regular class with an IAsyncEnumerable<StreamItem> ExecuteAsync(TCommand cmd, [EnumeratorCancellation] CancellationToken ct) method to stream data back to a Server-Sent Events endpoint. Because the handler needs services, I inject them into the endpoint and pass them in manually — which works, but prevents the handler class from using proper DI. Adding first-class streaming command support (similar to what MediatR offers) would be a great addition for FastEndpoints I think.

## Summary

- Adds `IStreamCommand<TResult>` / `IStreamCommandHandler<TCommand, TResult>` interfaces and 
  `StreamCommandHandler<TCommand, TResult>` base class, enabling commands that return 
  `IAsyncEnumerable<TResult>` via the command bus
- Adds generic stream command support via `RegisterGenericStreamCommand<TCommand, TResult, THandler>()`
- Adds `IStreamCommandMiddleware<TCommand, TResult>` pipeline support (mirrors the existing 
  `ICommandMiddleware` design), with `AddStreamCommandMiddleware()` registration and both 
  open-generic and closed-generic middleware registration
- Adds `RegisterTestStreamCommandHandler` for integration/unit test fake handler injection
- Refactors `CommandMiddlewareConfig` to inherit from `CommandMiddlewareConfigBase`, sharing 
  validation and DI registration logic with `StreamCommandMiddlewareConfig`
- Skips `IStreamCommand<>` types gracefully in `JobQueueExtensions` to prevent NRE

## Usage

```csharp
// define a stream command
public record GetNumbersCommand(int Count) : IStreamCommand<int>;

// implement a handler
public class GetNumbersHandler : IStreamCommandHandler<GetNumbersCommand, int>
{
    public async IAsyncEnumerable<int> ExecuteAsync(
        GetNumbersCommand cmd, 
        [EnumeratorCancellation] CancellationToken ct)
    {
        for (var i = 0; i < cmd.Count; i++)
        {
            await Task.Yield();
            yield return i;
        }
    }
}

// execute from an endpoint
await foreach (var item in new GetNumbersCommand(5).ExecuteAsync(ct))
    results.Add(item);


dj-nitehawk:
there were a few issues related native aot support and test handler registration. 
i believe they're sorted now.
thanks for the PR.
merging now!

PR #1107, opened 2026-05-18
Removed two redundant early-return guards in RequestBinder

SjorsGoering:
Removed duplicate early-return guards in BindUserClaims and BindHasPermissionProps.

The Count == 0 checks inside these methods were dead code — both methods are only called when _bindUserClaims / _bindPermissions is true, which is already gated on _fromClaimProps.Count > 0 and _hasPermissionProps.Count > 0 respectively in the constructor.

```cs
call sites:
_bindUserClaims = _fromClaimProps.Count > 0;
if (_bindUserClaims)
    BindUserClaims(req, ctx);

_bindPermissions = _hasPermissionProps.Count > 0;
if (_bindPermissions)
    BindHasPermissionProps(req, ctx);
```


dj-nitehawk:
good catch...
thank you!

PR #1110, opened 2026-05-21
Replaced runtime reflection in AccessControl Allow static constructor with source-generated alternative

SjorsGoering:
### What changed

The generated `Allow` class previously populated its `_permNames` and `_permCodes` dictionaries at
runtime by reflecting over all `public static` fields:

```csharp
foreach (var f in typeof(Allow).GetFields(BindingFlags.Public | BindingFlags.Static)
    .Where(t => !t.IsDefined(typeof(HideFromDocsAttribute))))
{
    var val = f.GetValue(null)?.ToString() ?? string.Empty;
    _permNames[f.Name] = val;
    _permCodes[val] = f.Name;
}
```
This is replaced by a source-generated static partial void InitPermissions() that populates the
dictionaries explicitly at compile time:

```csharp
static partial void InitPermissions()
{
    _permNames["MyPerm_Read"] = "ABC";
    _permCodes["ABC"] = "MyPerm_Read";
    // ...
}
```

An extra feature is that it is now possible to define a partial Allow class with partial string properties. The Source Generator will automatically duplicate the property definition and generate the code value.

```csharp
public static partial class Allow
{
    public static partial string Test_Partial_Permission { get; }
    // generated in Allow.g.cs: public static partial string Test_Partial_Permission { get => "KLE" };
}
```

The declaration and call live in the always-emitted Allow.b.g.cs; the implementation is emitted
in Allow.g.cs alongside the permission constants.

Why
AOT / IL trimming compatibility — reflection over static fields is not trim-safe and hinders Native AOT
Startup performance — no runtime field discovery; dictionaries are populated via direct assignments
Correctness — the generator already has complete knowledge of all permissions (endpoint-derived, const custom, and partial properties); there is no reason to rediscover them at runtime

Compatibility
Fully backwards compatible for all documented usage patterns:
- public const string fields, 
- AccessControl() endpoint definitions
- partial properties (added as extra feature),

The only edge case that does not work anymore is public static string (non-const, non-partial) fields in a partial Allow class. This scenario is currently ignored silently. Possible solutions
- generate an throw new NotSupportedException with a descriptive message telling the user what is not supported, mention this also in the docs and changelog
- Add support for it in the Source Generator


dj-nitehawk:
made some improvements and merged. thanks!

PR #1111, opened 2026-05-21
fix: uri encode query parameters

YoussefElshemi:
One example is if its a DateTimeOffset, and when you generate the string for it, you have 2026-05-21 08:56:33.684+0100
when this is sent to the test client, the request fails to deserialize because the + is treated as space, so it really gets

2026-05-21 08:56:33.684 0100, which then fails


dj-nitehawk:
thank you!

PR #1113, opened 2026-05-26
Add source generation registration support to FastEndpoints.Messaging and FastEndpoints.JobQueues

SjorsGoering:
# Add source generation and auto-registration support to FastEndpoints.Messaging via `[ModuleInitializer]`

## Summary

Previously:
- projects using `FastEndpoints.Messaging` made use of reflection-based types discovery at startup
- projects using `FastEndpoints.Generator` had to manually wire up discovered types at startup:

```csharp
services.AddFastEndpoints(o => o.SourceGeneratorDiscoveredTypes = DiscoveredTypes.All);
```

This was error-prone in multi-project solutions (easy to forget an assembly). This PR eliminates that step.

## How it works

A new `DiscoveredTypeRegistry` is introduced in `FastEndpoints.Core`. Both `DiscoveredTypesGenerator` and `GenericProcessorTypesGenerator` now emit a `[ModuleInitializer]` method in each generated class that automatically pushes the discovered types into this registry before `Main()` runs:

```csharp
[ModuleInitializer]
internal static void AutoRegister()
    => FastEndpoints.DiscoveredTypeRegistry.Register(All);
```

`AddFastEndpoints()` and `AddMessaging()` check `DiscoveredTypeRegistry` first and fall back to reflection-based assembly scanning only if the registry is empty. A conflict guard throws if both the generator and `EndpointDiscoveryOptions` (the reflection path) are configured simultaneously.

`EndpointDiscoveryOptions.SourceGeneratorDiscoveredTypes` is removed — it is no longer needed.

## Changes

- **`Src/Core/DiscoveredTypeRegistry.cs`** — new global registry with `Register`, `Override` (test helper), `HasTypes`, and `All`
- **`Src/Generator/DiscoveredTypesGenerator.cs`** — emits `[ModuleInitializer] AutoRegister()`
- **`Src/Generator/GenericProcessorTypesGenerator.cs`** — same
- **`Src/Library/Main/EndpointDiscoveryOptions.cs`** — removes `SourceGeneratorDiscoveredTypes`
- **`Src/Library/Main/MainExtensions.cs`** — `EndpointDiscoveryOptions` is now optional (`null` when not provided)
- **`Src/Library/Main/EndpointData.cs`** — checks registry before falling back to `AssemblyScanner`
- **`Src/Messaging/Messaging/MessagingExtensions.cs`** — same pattern; adds conflict guard
- **`Src/Core/FastEndpoints.Core.csproj`** — adds `InternalsVisibleTo` for `Int.Agents` test project
- **Tests** — all usages of `SourceGeneratorDiscoveredTypes` replaced with `DiscoveredTypeRegistry.Override(...)`; tests that need pure reflection call `DiscoveredTypeRegistry.Override()` (no args) to clear the registry

## Before / After

**Before:**
```csharp
// Each project using FastEndpoints.Generator must add the DiscoveredTypes manually
services.AddFastEndpoints(o => o.SourceGeneratorDiscoveredTypes = DiscoveredTypes.All);
```

**After:**
```csharp
// nothing needed — the generator handles it automatically
services.AddFastEndpoints();
```


dj-nitehawk:
Not merge-ready.                                                                                                                                                                                      
                                                                                                                                                                                                       
 Main issue: the auto-registration only works for assemblies that are already loaded.                                                                                                                  
                                                                                                                                                                                                       
 [ModuleInitializer] runs when an assembly loads, but AddFastEndpoints() / AddMessaging() read DiscoveredTypeRegistry immediately. Referenced endpoint/handler projects are not guaranteed to be       
 loaded at that point, so their generated registrations may never be present.                                                                                                                          
                                                                                                                                                                                                       
 Observed failure:                                                                                                                                                                                     
                                                                                                                                                                                                       
 - Web app references an endpoint class library.                                                                                                                                                       
 - Library uses FastEndpoints.Generator.                                                                                                                                                               
 - App calls only builder.Services.AddFastEndpoints().                                                                                                                                                 
 - Startup fails with:                                                                                                                                                                                 
                                                                                                                                                                                                       
```
FastEndpoints was unable to find any endpoint declarations! 
```                                                                                                                                                                                                   
                                                                                                                                                                                                       
 If a type from the library is explicitly touched before AddFastEndpoints(), it succeeds, confirming the assembly-loading issue.                                                                       
                                                                                                                                                                                                       
 Second issue: registry discovery is all-or-nothing.                                                                                                                                                   
                                                                                                                                                                                                       
 If DiscoveredTypeRegistry.HasTypes is true, reflection scanning is skipped entirely. This can cause silent partial discovery: types from already-loaded generated assemblies are found, while         
 endpoints/handlers in unloaded referenced assemblies are missed.                                                                                                                                      
                                                                                                                                                                                                       
 Third issue: the conflict guards block workarounds.                                                                                                                                                   
                                                                                                                                                                                                       
 AddFastEndpoints(options => ...) and AddMessaging(assemblies) throw when generated types already exist in the registry. That prevents users from explicitly passing assemblies to recover missing     
 types.                                                                                                                                                                                                
                                                                                                                                                                                                       
 Tests/build pass, but the core multi-project scenario the PR aims to solve is not reliable.

SjorsGoering:
Good catch. The ModuleInitializer trick will indeed not work here. I've reverted those related changes.

What's left in this PR are 2 things:
- When using the Messaging and/or JobQueue packages standalone, it is now possible to use the source-generated type discovery feature from the Generator package.
- I've separated the AddFastEndpoints registration call into 2 overloads — one for source-generated and one for reflection-based type discovery — since I noticed that EndpointDiscoveryOptions had 1 property for source-generator discovery while all other properties were only used in the reflection-based discovery code path. With this change it is no longer possible to accidentally mix both functionalities.
- Changelog has been updated to mention that public static partial string properties declared in a partial Allow class no longer need a manual implementation (feature added in #1110).

Let me know what you think @dj-nitehawk

dj-nitehawk:
Not merge-ready yet. Please address the following issues.

## 1. Fix `AddMessaging()` overload ambiguity

**File:** `Src/Messaging/Messaging/MessagingExtensions.cs`

Adding `params List<Type>[] discoveredTypes` makes existing calls to:

```csharp
services.AddMessaging();
```

ambiguous with the existing reflection-based overload:

```csharp
AddMessaging(IServiceCollection, params Assembly[]?)
AddMessaging(IServiceCollection, params List<Type>[])
```

Verified with a temp consumer build: `CS0121` between the two `AddMessaging` overloads.

## 2. Avoid reflection scanning in the source-generated JobQueues path

**Files:**
- `Src/JobQueues/JobQueueExtensions.cs`
- `Src/Messaging/Messaging/MessagingExtensions.cs`
- `Src/Core/AssemblyScanner/AssemblyScanner.cs`

This documented pattern still triggers reflection scanning:

```csharp
builder.Services
       .AddFastEndpoints(DiscoveredTypes.All)
       .AddJobQueues<Job, JobStorage>();
```

`AddJobQueues<TStorageRecord,TStorageProvider>()` calls `AddMessaging((Assembly[]?)null)`, which immediately runs `AssemblyScanner.ScanForTypes(...)`. This contradicts the source-generated/AOT path documented in `Src/Library/changelog.md` and used by `TestHarness/NativeAotChecker/Program.cs`.

## 3. Fix `AddFastEndpoints(null)` ambiguity

**File:** `Src/Library/Main/MainExtensions.cs`

Explicit-null callers now hit ambiguity between:

```csharp
AddFastEndpoints(IServiceCollection, Action<EndpointDiscoveryOptions>?)
AddFastEndpoints(IServiceCollection, params List<Type>[])
```

SjorsGoering:
I've pushed some fixes, can you have another look @dj-nitehawk

dj-nitehawk:
looks good now. thanks!

Issue #1112, opened 2026-05-22
_childAdaptorValidators -System.InvalidOperationException: Operations that change non-concurrent collections must have exclusive access


vjohansen:
I get an exception which i think is from

https://github.com/FastEndpoints/FastEndpoints/blob/abc2be3a7142c394fcd4c10d7061772ca9cb64e1/Src/Swagger/ValidationSchemaProcessor.cs#L202

The actual exception (ApplyValidationRule)

```
Exception while processing "ProductionUnitCreationRequestValidator" [FastEndpoints.Swagger.ValidationSchemaProcessor]
System.InvalidOperationException: Operations that change non-concurrent collections must have exclusive access. A concurrent update was performed on this collection and corrupted its state. The collection's state is no longer correct.
   at System.Collections.Generic.Dictionary`2.set_Item(TKey key, TValue value)
   at FastEndpoints.Swagger.ValidationSchemaProcessor.ApplyValidationRule(JsonSchema schema, IValidationRule validationRule, String propertyName, IServiceProvider services)
   at FastEndpoints.Swagger.ValidationSchemaProcessor.TryApplyValidation(JsonSchema schema, ReadOnlyDictionary`2 rulesDict, String propertyName, String parameterPrefix, IServiceProvider services)
   at FastEndpoints.Swagger.ValidationSchemaProcessor.ApplyRulesToSchema(JsonSchema schema, ReadOnlyDictionary`2 rulesDict, String propertyPrefix, IServiceProvider services)
   at FastEndpoints.Swagger.ValidationSchemaProcessor.ApplyValidator(JsonSchema schema, IValidator validator, String propertyPrefix, IServiceProvider services)
   at FastEndpoints.Swagger.ValidationSchemaProcessor.Process(SchemaProcessorContext context)
```

My code looks like this

```
	public ProductionUnitCreationRequestValidator() => RuleFor(x => x.Unit)
		.SetValidator((req, unit) => new ProductionUnitRequestDtoValidator(req.TenantId, null));
```

Maybe `_childAdaptorValidators` should use ConcurrentDictionary?

dj-nitehawk:
will look in to this as soon as i can.
in the meantime, give the new `FastEndpoints.OpenApi` package a try.
a lot of the issues with `FastEndpoints.Swagger` has been solved.
see https://github.com/FastEndpoints/FastEndpoints/issues/1094#issue-4355324264 for more info.

SjorsGoering:
[...]

Issue #1109, opened 2026-05-20
Swagger: nested SetValidator constraints dropped when generating multiple versioned documents in same process


sdukehart-omnesoft:
## Summary

When two `SwaggerDocument` registrations share a nested DTO whose validator
uses `RuleForEach(...).SetValidator(new XValidator())`, whichever document
is generated SECOND in a given process loses the nested validator's
property-level constraints (`minimum`, `minLength`, `maxLength`, `required`
fields) on the shared `$ref`'d schema component.

The bug is symmetric — order is what matters, not which version. The first
document is correct; the second loses constraints. State persists for the
process lifetime.

## Minimal Repro

`FeBugRepro.csproj`:

```xml
<Project Sdk="Microsoft.NET.Sdk.Web">
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net10.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="FastEndpoints" Version="8.1.0" />
    <PackageReference Include="FastEndpoints.Swagger" Version="8.1.0" />
  </ItemGroup>
</Project>
```

`Program.cs`:

```csharp
using FastEndpoints;
using FastEndpoints.Swagger;
using FluentValidation;

var builder = WebApplication.CreateBuilder(args);
builder.Services.AddFastEndpoints();
builder.Services.SwaggerDocument(o =>
{
    o.MinEndpointVersion = 1;
    o.MaxEndpointVersion = 1;
    o.DocumentSettings = s => { s.DocumentName = "v1"; s.Title = "V1"; s.Version = "v1"; };
});
builder.Services.SwaggerDocument(o =>
{
    o.MinEndpointVersion = 2;
    o.MaxEndpointVersion = 2;
    o.DocumentSettings = s => { s.DocumentName = "v2"; s.Title = "V2"; s.Version = "v2"; };
});

var app = builder.Build();
app.UseFastEndpoints(o =>
{
    o.Versioning.Prefix = "v";
    o.Versioning.PrependToRoute = true;
});
app.UseSwaggerGen();
app.Run();

public record OrderLine
{
    public required string ProductId { get; init; }
    public required decimal Quantity { get; init; }
}

public class OrderLineValidator : AbstractValidator<OrderLine>
{
    public OrderLineValidator()
    {
[...]

PR #1114, opened 2026-05-27
Fix: ValidationSchemaProcessor to prevent concurrent corruption and cross-document constraint loss

SjorsGoering:
# Fix for #1112 and #1109 _childAdaptorValidators -System.InvalidOperationException: Operations that change non-concurrent collections must have exclusive access and nested SetValidator constraints dropped when generating multiple versioned documents in same process

### Background

`ValidationSchemaProcessor` is registered as a **singleton** via `CreateSingleton`
(which uses `GetOrAdd` on an internal concurrent cache), so every `SwaggerDocument`
in the application shares the same processor instance. This exposed two distinct bugs
in the `_childAdaptorValidators` field — one concurrency issue and one document-lifetime
issue — both fixed in this PR.

---

### Bug 1 — Concurrent dictionary corruption (race condition)

**Symptom:** Hitting the swagger endpoint with simultaneous requests threw
`InvalidOperationException` with a corrupted `Dictionary` internal state.

**Root cause:** `_childAdaptorValidators` was declared as a plain
`Dictionary<string, IValidator>`. Multiple threads calling `Process()` concurrently
could both find a key absent and write to the dictionary simultaneously, corrupting
its internal state.

**Initial fix (commit dd18fe51):** Changed the field to
`ConcurrentDictionary<string, IValidator>` so concurrent reads and writes are safe.

---

### Bug 2 — Cross-document constraint loss (state leakage)

**Symptom:** When two `SwaggerDocument` registrations share a DTO whose validator
uses `RuleForEach(...).SetValidator(new ChildValidator())` (or
`RuleFor(...).SetValidator(...)`), only the **first** document generated in a given
process lifetime has the nested validator's constraints (`required`, `minLength`,
`minimum`, etc.) on the shared schema component. The second document's copy of that
schema is emitted with all constraints stripped. The effect is symmetric — whichever
document is requested first wins; the other loses.

**Root cause:** The `_childAdaptorValidators` cache (even as `ConcurrentDictionary`)
accumulated validator-type entries across all document passes for the lifetime of the
process. On the second document's pass, `TryGetValue` succeeded for types already seen
in the first pass, causing the code to `continue` and skip `ApplyValidator` entirely —
leaving the second document's schemas without any nested constraints.

**Reproduction:**

```csharp
// Both validators use the same OrderLineValidator
public class CreateOrderV1Validator : Validator<CreateOrderV1Request>
{
    public CreateOrderV1Validator()
        => RuleForEach(x => x.Lines).SetValidator(new OrderLineValidator());
}
public class CreateOrderV2Validator : Validator<CreateOrderV2Request>
{
    public CreateOrderV2Validator()
        => RuleForEach(x => x.Lines).SetValidator(new OrderLineValidator());
}
```
Fetch v1 first, then v2:

v1/swagger.json → OrderLine has required, minLength, minimum ✓
v2/swagger.json → OrderLine has no constraints ✗
Combined fix
Both bugs share the same root cause: _childAdaptorValidators as a long-lived
instance field. The complete solution is to remove the field entirely and replace
it with a fresh Dictionary<string, IValidator> created at the start of each
Process() invocation, then threaded through the private call chain:

Process()  ← creates dictionary here; discarded on return
  └─ ApplyValidator(…, dict)
       ├─ ApplyRulesToSchema(…, dict)
       │    └─ TryApplyValidation(…, dict)
       │         └─ ApplyValidationRule(…, dict)
       │              └─ ApplyValidator (recursive, same dict)
       └─ ApplyRulesFromIncludedValidators(…, dict)
This supersedes the ConcurrentDictionary workaround:

Concurrency — a call-local Dictionary is never shared between threads, so there is no concurrent write to protect against.
Cross-document leakage — the dict is discarded when Process() returns, so each document pass starts clean and applies all child-validator constraints to its own fresh schema objects.
Recursion guard — still works correctly: within one Process() call the dict accumulates seen validator types as before, preventing infinite loops from self-referential validators.
No logic changes — only the dictionary's lifetime narrows from process-scoped to
per-Process()-call.

Test
Added MultiDocSharedOrderLine + MultiDocSharedOrderLineValidator
(using RuleForEach + SetValidator) along with two swagger-review endpoints
(MultiDocOrderEndpointA, MultiDocOrderEndpointB) that both reference the same
child validator. The regression test
shared_nested_validator_constraints_survive_multi_document_generation generates
"Swagger Review" then "Swagger Review Empty Schema" and asserts that
sku.minLength and qty.exclusiveMinimum are present in both documents.

Note: The Int.Swagger integration test project is (and was already) failing before these changes because of IOpenApiDocumentGenerator not being registered. Solving this is probably not worth it since the Swagger package is about to be phased out.

### DIFF
--- a/FastEndpoints.slnx
+++ b/FastEndpoints.slnx
@@ -29,10 +29,10 @@
         <Project Path="Src/ClientGen/FastEndpoints.ClientGen.csproj"/>
         <Project Path="Src/Swagger/FastEndpoints.Swagger.csproj"/>
     </Folder>
-    <Folder Name="/Legacy/Tests/">
-        <Project Path="Tests/IntegrationTests/FastEndpoints.Swagger/Int.Swagger.csproj"/>
-        <Project Path="Tests/UnitTests/FastEndpoints.Swagger/Unit.Swagger.csproj"/>
-    </Folder>
+    <!-- <Folder Name="/Legacy/Tests/"> -->
+    <!--     <Project Path="Tests/IntegrationTests/FastEndpoints.Swagger/Int.Swagger.csproj"/> -->
+    <!--     <Project Path="Tests/UnitTests/FastEndpoints.Swagger/Unit.Swagger.csproj"/> -->
+    <!-- </Folder> -->
     <Folder Name="/TestHarnesses/">
         <Project Path="TestHarness/OData/OData.csproj"/>
         <Project Path="TestHarness/OpenApi.Kiota/OpenApi.Kiota.csproj"/>
@@ -44,8 +44,8 @@
     </Folder>
     <Folder Name="/Tests/IntegrationTests/">
         <Project Path="Tests/IntegrationTests/FastEndpoints.Agents/Int.Agents.csproj"/>
-        <!-- <Project Path="Tests/IntegrationTests/FastEndpoints.OpenApi.Kiota/Int.OpenApi.Kiota.csproj"/> -->
-        <!-- <Project Path="Tests/IntegrationTests/FastEndpoints.OpenApi/Int.OpenApi.csproj"/> -->
+        <Project Path="Tests/IntegrationTests/FastEndpoints.OpenApi.Kiota/Int.OpenApi.Kiota.csproj"/>
+        <Project Path="Tests/IntegrationTests/FastEndpoints.OpenApi/Int.OpenApi.csproj"/>
         <Project Path="Tests/IntegrationTests/FastEndpoints/Int.FastEndpoints.csproj"/>
         <Project Path="Tests/IntegrationTests/FastEndpoints.OData/Int.OData.csproj"/>
     </Folder>
--- a/FastEndpoints.slnx
+++ b/FastEndpoints.slnx
@@ -29,10 +29,10 @@
         <Project Path="Src/ClientGen/FastEndpoints.ClientGen.csproj"/>
         <Project Path="Src/Swagger/FastEndpoints.Swagger.csproj"/>
     </Folder>
-    <!-- <Folder Name="/Legacy/Tests/"> -->
-    <!--     <Project Path="Tests/IntegrationTests/FastEndpoints.Swagger/Int.Swagger.csproj"/> -->
-    <!--     <Project Path="Tests/UnitTests/FastEndpoints.Swagger/Unit.Swagger.csproj"/> -->
-    <!-- </Folder> -->
+    <Folder Name="/Legacy/Tests/">
+        <Project Path="Tests/IntegrationTests/FastEndpoints.Swagger/Int.Swagger.csproj"/>
+        <Project Path="Tests/UnitTests/FastEndpoints.Swagger/Unit.Swagger.csproj"/>
+    </Folder>
     <Folder Name="/TestHarnesses/">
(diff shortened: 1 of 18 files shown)


dj-nitehawk:
Status: **not merge-ready yet**.

Remaining issues to fix before merge:

## 1. Use an active recursion stack, not a per-process-call “seen” dictionary

Current code still treats a child validator type as already handled for the rest of the current `Process()` call:

```csharp
if (!validatorType.IsInterface &&
    !childAdaptorValidators.TryGetValue(validatorType.FullName!, out var childValidator))
{
    childValidator = childAdaptorValidators[validatorType.FullName!] =
        (IValidator)_serviceResolver.CreateInstance(validatorType, services);
}
else
    continue;
```

This fixes cross-document reuse, but can still skip legitimate repeated uses of the same child validator type in one request/schema traversal, e.g.:

```csharp
RuleFor(x => x.BillingAddress).SetValidator(new AddressValidator());
RuleFor(x => x.ShippingAddress).SetValidator(new AddressValidator());
```

Recommended fix:

- Replace the “seen forever in this `Process()` call” dictionary behavior with an active recursion stack.
- Use `HashSet<Type> activeChildValidators`.
- `Add()` before recursively applying a child validator.
- `Remove()` in a `finally` after recursion returns.
- If instance caching is still desired, keep it separate from recursion detection.

The newer OpenAPI implementation already follows this safer pattern in `Src/OpenApi/Validation/ChildValidatorResolver.cs`.

## 2. Add a regression test for repeated same validator type in one schema

Add a legacy Swagger test where one request has two different properties using the same child validator type. Assert both generated child schemas/properties get the expected constraints.

Example shape:

```csharp
sealed class Address
{
    public string Zip { get; init; } = string.Empty;
}

sealed class AddressValidator : Validator<Address>
{
    public AddressValidator() => RuleFor(x => x.Zip).NotEmpty();
}

sealed class Request
{
    public Address BillingAddress { get; init; } = new();
    public Address ShippingAddress { get; init; } = new();
}

sealed class RequestValidator : Validator<Request>
{
    public RequestValidator()
    {
        RuleFor(x => x.BillingAddress).SetValidator(new AddressValidator());
        RuleFor(x => x.ShippingAddress).SetValidator(new AddressValidator());
    }
}
```

Expected: both address schemas/properties receive `zip` constraints.

## 3. Add a true concurrency regression test

The current added test verifies sequential multi-document generation. Good, but it does not exercise concurrent generation.

Add a focused test using `Task.WhenAll(...)`, e.g.:

```csharp
var docs = await Task.WhenAll(
    App.DocGenerator.GenerateAsync("Swagger Review"),
    App.DocGenerator.GenerateAsync("Swagger Review Empty Schema"),
    App.DocGenerator.GenerateAsync("Release 2.0"));
```

Then assert the nested validator constraints are present in each relevant generated document.

## 4. Minor changelog wording

`Src/Library/changelog.md` says the dictionary is scoped to each `ProcessAsync` call, but the actual method is `Process()`.

Change `ProcessAsync` → `Process`.

---

OpenAPI test disabling for legacy Swagger work should not be treated as a blocker.

SjorsGoering:
I applied the suggestions, let me know what you think @dj-nitehawk

dj-nitehawk:
all good now.
thank you!

PR #1115, opened 2026-05-29
Fix: configure serializer options once per process to prevent concurrent JsonSerializerOptions read-only race

ebarkalov:
## Background

`FastEndpoints.Config` stores serializer settings in a **process-global static**. `Config.Serializer.Options` (`Cfg.SerOpts.Options`) is a single `System.Text.Json.JsonSerializerOptions` instance shared by every host in the process:

```csharp
// Src/Library/Config/Config.cs
internal static readonly SerializerOptions SerOpts = new();

// Src/Library/Config/SerializerOptions.cs
public JsonSerializerOptions Options { get; internal set; } = new();
```

This is fine for the normal "one app per process" scenario. It becomes a problem when **multiple hosts are created in the same process** — the typical setup for `WebApplicationFactory`-based integration tests, **especially when the test framework runs test classes in parallel** (e.g. TUnit, which is parallel-by-default, or xUnit without disabling parallelization). Each host's startup calls `MapFastEndpoints`, which **re-points and mutates that shared static**.

## Symptom

Running such integration tests in parallel intermittently fails at host startup with:

```
System.InvalidOperationException: This JsonSerializerOptions instance is read-only or has already been used in serialization or deserialization.
   at System.Text.Json.ThrowHelper.ThrowInvalidOperationException_SerializerOptionsReadOnly(JsonSerializerContext context)
   at System.Text.Json.JsonSerializerOptions.set_DefaultIgnoreCondition(JsonIgnoreCondition value)
   at <the user's UseFastEndpoints config action>          // e.g. cfg.Serializer.Options.DefaultIgnoreCondition = ...
   at FastEndpoints.MainExtensions.MapFastEndpoints(IEndpointRouteBuilder app, Action`1 configAction)
   at FastEndpoints.MainExtensions.UseFastEndpoints(IApplicationBuilder app, Action`1 configAction)
   at Program.<Main>$(String[] args)
   ...
   at Microsoft.AspNetCore.Mvc.Testing.WebApplicationFactory`1.CreateHost(IHostBuilder builder)
   at Microsoft.AspNetCore.Mvc.Testing.WebApplicationFactory`1.StartServer()
```

It is **sporadic** and **timing-dependent** — it shows up more often on CI, on Linux, and on higher core counts (more parallelism), and frequently passes locally. The throwing statement varies: it can be any mutation of the options (e.g. `set_DefaultIgnoreCondition`, `set_PropertyNamingPolicy`, `Converters.Add(...)`), or one of FastEndpoints' own internal mutations (see below).

## Root cause

`MapFastEndpoints` configures the shared static in place:

```csharp
// Src/Library/Main/MainExtensions.cs (before)
var serializerOptions = app.ServiceProvider.GetService<IOptions<JsonOptions>>()?.Value.SerializerOptions;

if (serializerOptions is not null)
    Cfg.SerOpts.Options = new(serializerOptions);                       // (1) publish a fresh instance to the shared static

Cfg.SerOpts.Options.ConfigureSerializer(                                // (2) then mutate it in place
    app.ServiceProvider.GetRequiredService<Cfg>(), configAction);
```

…and `ConfigureSerializer` mutates it further, including invoking the user's config action:

```csharp
// Src/Library/Config/ConfigExtensions.cs
opts.TypeInfoResolver = opts.TypeInfoResolver?.WithAddedModifier(...);                 // mutate
configAction?.Invoke(cfg);                                                            // mutate (cfg.Serializer.Options.X = ...)
opts.TypeInfoResolverChain.Insert(0, new FastEndpointsSerializerContext(new(opts)));  // mutate
```

A `JsonSerializerOptions` becomes **read-only on its first (de)serialization** (FastEndpoints' `RequestDeserializer` / `ResponseSerializer` call `ReadFromJsonAsync` / `WriteAsJsonAsync` with `SerOpts.Options`). With two hosts (A and B) starting in the same process, this interleaving occurs:

```
Host B startup:  Cfg.SerOpts.Options = newB            // (1) re-point the shared static
Host A serving:  reads Cfg.SerOpts.Options (== newB)   // serializes a request/response
                 -> first use freezes newB             // newB is now read-only
Host B startup:  newB.set_DefaultIgnoreCondition(...)  // (2) mutate -> THROWS (read-only)
```

Because the shared static is re-pointed by B and then read by A's request pipeline, **A freezes the very instance B is still configuring**.

Importantly, this cannot be worked around from the application side:

- Even a **no-op** config action still throws, because FastEndpoints itself mutates the shared instance (`TypeInfoResolver` / `TypeInfoResolverChain.Insert`) during every host's startup.
- A **user-level lock** around `UseFastEndpoints` does not help either — the freeze happens during request **serving**, which is not (and cannot be) held under that lock.

## Fix

Configure the process-global serializer options **exactly once**, under a lock (double-checked):

```csharp
static readonly object _serializerConfigLock = new();
static volatile bool _serializerConfigured;

static void ConfigureSerializerOnce(IEndpointRouteBuilder app, Action<Cfg>? configAction)
{
    if (_serializerConfigured)
        return;

    lock (_serializerConfigLock)
    {
        if (_serializerConfigured)
            return;

        var serializerOptions = app.ServiceProvider.GetService<IOptions<JsonOptions>>()?.Value.SerializerOptions;

        if (serializerOptions is not null)
            Cfg.SerOpts.Options = new(serializerOptions);

        Cfg.SerOpts.Options.ConfigureSerializer(app.ServiceProvider.GetRequiredService<Cfg>(), configAction);

        _serializerConfigured = true;
    }
}
```

`MapFastEndpoints` now calls `ConfigureSerializerOnce(app, configAction)` in place of the inline configuration block.

### Why this is correct and race-free

- The single configuration provably runs **before any host can serve**: every host must pass through `MapFastEndpoints` (and therefore this gate) before its request pipeline is live. The first host to take the lock configures the options while **no other host has finished startup**, so no request pipeline can be serializing/freezing the instance at that moment.
- Every subsequent host **skips** the re-point/reconfigure and reuses the already-configured instance. Once the first (de)serialization freezes it, it is never mutated again — so the read-only mutation can no longer occur.

### Non-breaking / parallelism preserved

- **No behavior change** for the normal single-host-per-process case: there is exactly one `MapFastEndpoints` call, so the options are configured once, exactly as before.
- Hosts still **start up and serve fully in parallel** — only the one-time options setup is synchronized (an uncontended `volatile` read on the fast path after the first call). It does **not** force tests to run sequentially.
- This keeps the existing "global configuration per process" model; it just makes that one-time setup deterministic and thread-safe.

This eliminated a long-standing sporadic CI failure in a parallel `WebApplicationFactory` (TUnit) integration-test suite, while keeping the tests running in parallel.

### DIFF
--- a/Src/Library/Endpoint/Idempotency/IdempotencyOptions.cs
+++ b/Src/Library/Endpoint/Idempotency/IdempotencyOptions.cs
@@ -1,12 +1,5 @@
 using Microsoft.Net.Http.Headers;
 
-#if NET9_0_OR_GREATER
-using Lock = System.Threading.Lock;
-
-#else
-    using Lock = object;
-#endif
-
 namespace FastEndpoints;
 
 /// <summary>
--- a/Src/Library/Main/MainExtensions.cs
+++ b/Src/Library/Main/MainExtensions.cs
@@ -66,13 +66,12 @@ public static class MainExtensions
         return app;
     }
 
-    static readonly object _serializerConfigLock = new();
+    static readonly Lock _serializerConfigLock = new();
     static volatile bool _serializerConfigured;
 
     public static IEndpointRouteBuilder MapFastEndpoints(this IEndpointRouteBuilder app, Action<Cfg>? configAction = null)
     {
         ServiceResolver.Instance = app.ServiceProvider.GetRequiredService<IServiceResolver>();
-
         ConfigureSerializerOnce(app, configAction);
         Cfg.BndOpts.AddTypedHeaderValueParsers();
 
@@ -212,14 +211,8 @@ public static class MainExtensions
         return app;
     }
 
-    /// <summary>
-    ///     Configures the process-wide serializer options (<see cref="SerializerOptions.Options" />) exactly once.
-    /// </summary>
     static void ConfigureSerializerOnce(IEndpointRouteBuilder app, Action<Cfg>? configAction)
     {
-        if (_serializerConfigured)
-            return;
-
         lock (_serializerConfigLock)
         {
             if (_serializerConfigured)
--- a/Src/Library/Main/MainExtensions.cs
+++ b/Src/Library/Main/MainExtensions.cs
@@ -66,16 +66,14 @@ public static class MainExtensions
         return app;
     }
 
+    static readonly object _serializerConfigLock = new();
+    static volatile bool _serializerConfigured;
+
     public static IEndpointRouteBuilder MapFastEndpoints(this IEndpointRouteBuilder app, Action<Cfg>? configAction = null)
     {
         ServiceResolver.Instance = app.ServiceProvider.GetRequiredService<IServiceResolver>();
 
-        var serializerOptions = app.ServiceProvider.GetService<IOptions<JsonOptions>>()?.Value.SerializerOptions;
-
-        if (serializerOptions is not null)
-            Cfg.SerOpts.Options = new(serializerOptions);
-
-        Cfg.SerOpts.Options.ConfigureSerializer(app.ServiceProvider.GetRequiredService<Cfg>(), configAction);
+        ConfigureSerializerOnce(app, configAction);
         Cfg.BndOpts.AddTypedHeaderValueParsers();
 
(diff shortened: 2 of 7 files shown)


dj-nitehawk:
merged! thank you!

PR #1117, opened 2026-06-02
Source-generate [KeyedService] keys into ReflectionCache for property injection

SjorsGoering:
### Problem

`GetServiceBoundEpProps()` always used runtime reflection to discover `[KeyedService]` attributes on endpoint properties, even in assemblies that opted into source generation via `ReflectionCache`. This meant keyed-service injection bypassed the source-generated cache entirely and could not benefit from AOT-safe, allocation-free property resolution.

### Solution

**`PropertyDefinition`** gets a new `ServiceKey` property, making it the single place to carry keyed-service metadata for a property — whether populated by the generator or set manually.

**`ReflectionGenerator`** now emits `ServiceKey = "..."` into the generated `PropertyDefinition` initializer for every endpoint property decorated with `[KeyedService]`. The `Prop` struct gains an `isEndpointClass` flag so key extraction only happens when processing endpoint classes (not request DTOs).

**`GetServiceBoundEpProps()`** checks the `ReflectionCache` first. If the endpoint type has a cached `TypeDefinition` with properties, it reads `ServiceKey` directly from `PropertyDefinition` — no reflection. The runtime-reflection fallback (reading
`[KeyedServiceAttribute]` via `GetCustomAttribute`) is kept for assemblies without source generation.

### Test

A new unit test covers both code paths in isolation:
- **Reflection path** — endpoint type absent from the cache; keys are read from   `[KeyedService]` attributes via reflection.
- **Source-gen path** — cache pre-seeded with `PropertyDefinition.ServiceKey` values   (as the generator would emit); verifies those values are used instead of reflection.


dj-nitehawk:
## Required fixes

### 1. Generate endpoint property metadata independently of request DTO metadata

Current generator only creates the endpoint-class cache entry when request DTO processing reaches `Properties.Count > 0`. This skips common cases such as:

- `EndpointWithoutRequest<TResponse>` / `EmptyRequest`
- request DTOs with zero bindable properties
- blacklisted request DTO types

Fix `ReflectionGenerator.TypeInfo` so endpoint property-injection metadata is collected from the endpoint type itself regardless of the request DTO outcome.

Expected behavior:

```csharp
sealed class MyEndpoint : EndpointWithoutRequest<string>
{
    [KeyedService("A")]
    public IMyService Service { get; set; } = default!;
}
```

After generated registration:

```csharp
Config.BndOpts.ReflectionCache.AddFromMyAssembly();
```

The cache should contain `typeof(MyEndpoint)` with `ServiceKey == "A"` for `Service`.

### 2. Escape generated service key literals

Do not emit raw key strings like this:

```csharp
$", ServiceKey = \"{p.ServiceKey}\""
```

Use Roslyn literal formatting instead:

```csharp
SymbolDisplay.FormatLiteral(p.ServiceKey, quote: true)
```

Then emit:

```csharp
$", ServiceKey = {SymbolDisplay.FormatLiteral(p.ServiceKey, quote: true)}"
```

This prevents broken generated code for valid keys containing quotes, backslashes, newlines, etc.

### 3. Add real source-generator coverage

The current unit test manually inserts `PropertyDefinition.ServiceKey` into `ReflectionCache`. Keep that if useful, but add tests proving the generator emits the key.

Cover at least:

- `EndpointWithoutRequest<TResponse>` with `[KeyedService]`
- endpoint with a request DTO that has no bindable properties
- keyed service values that require escaping, e.g. `tenant\"a` and `tenant\\path`

The test should call the generated `AddFrom...()` method or compile source with `ReflectionGenerator` and assert the generated cache content/source.

### 4. Confirm the existing keyed-service endpoint uses the generated path

Use the existing test harness endpoint:

```text
TestHarness/Web/[Features]/TestCases/DependencyInjection/KeyedServicesTests/KeyedService.cs
```

After `Web.GeneratedReflection.AddFromWeb(cache)`, `cache` should contain the keyed service endpoint type and the `KeyedService` property should have `ServiceKey == "AAA"`.

dj-nitehawk:
Not merge-ready yet.

## 1. Preserve `[KeyedService]` when cache lacks `ServiceKey`

`EndpointDefinition.GetServiceBoundEpProps()` currently trusts cached `PropertyDefinition.ServiceKey` directly.

Update the cached path to fall back to the actual property attribute when `ServiceKey` is null:

```csharp
kv.Value.ServiceKey ?? kv.Key.GetCustomAttribute<KeyedServiceAttribute>()?.Key
```

Target file:

- `Src/Library/Endpoint/Auxiliary/EndpointDefinition.cs`

Add a unit test where:

- endpoint property has `[KeyedService("KEY")]`
- `ReflectionCache` already contains that endpoint property
- cached `PropertyDefinition.ServiceKey` is null
- `ServiceBoundEpProps` still returns `"KEY"`

## 2. Emit `ServiceKey` for `init` endpoint properties

The generator currently emits plain `new()` for init-only properties, dropping captured `ServiceKey`.

Fix rendering so this works:

```csharp
[KeyedService("KEY")]
public IService Service { get; init; } = default!;
```

Expected generated metadata must include:

```csharp
ServiceKey = "KEY"
```

Target file:

- `Src/Generator/ReflectionGenerator.cs`

Add a generator test covering a keyed `init` endpoint property.

## Merge criteria

Ready after:

- cached reflection path preserves attribute keys when `ServiceKey` is absent
- source-generated path preserves keys for `init` properties
- tests cover both regressions

SjorsGoering:
I've made an attempt to push some changes to fix both issues, you can have another look @dj-nitehawk .

dj-nitehawk:
all good now.
thank you!

PR #1118, opened 2026-06-04
Fix ObjectDisposedException from ServiceResolver root-provider fallback during requests

ebarkalov:
## Problem

In multi-host test processes (several `WebApplicationFactory` instances in one test run), command execution can throw:

```
System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'IServiceProvider'.
   at ...ServiceProviderEngineScope.GetService(Type serviceType)
   at FastEndpoints.ServiceResolver.TryResolve(Type typeOfService)
   at FastEndpoints.CommandExtensions.PrepareExecution[TResult](...)
   at FastEndpoints.CommandExtensions.ExecuteAsync[TResult](ICommand`1 command, CancellationToken ct)
```

It surfaced after upgrading to `8.2.0-beta.*` and is intermittent (depends on another test host being torn down during a test's lifetime).

## Root cause

Since #1017 (command/event bus extraction), bootstrap moved from `Cfg.ServiceResolver` to the process-global static `ServiceResolver.Instance`, and the command bus now resolves through that static. On every `MapFastEndpoints()` call:

- `ServiceResolver.Instance` is overwritten with the current host's resolver (capturing that host's **root** `IServiceProvider`), and
- `CommandExtensions.TestCommandHandlerMarker` is set unconditionally.

So `CommandExtensions.PrepareExecution` always runs `resolver.TryResolve(TestCommandHandlerMarker)`, and `ServiceResolver.TryResolve` falls back to the captured root `provider` when the request scope doesn't contain the type:

```csharp
public object? TryResolve(Type typeOfService)
    => ctxAccessor?.HttpContext?.RequestServices.GetService(typeOfService) ??
       provider.GetService(typeOfService);
```

With multiple hosts in one process, `Instance` points at whichever host bootstrapped last. Once that host is disposed, `Instance` still references it, and the marker type (not registered in the live request scope) makes `TryResolve` fall back to the **disposed** root provider → `ObjectDisposedException`.

## Fix

When a request scope is active, resolve solely from `HttpContext.RequestServices` instead of falling back to the captured root `provider`. The request scope is a child of the root, so it already resolves every registered service; the `?? provider` fallback only ever returned `null` for unregistered services anyway. The root is still used when there is no active request scope (background command/event execution, unit-test setup).

Applied to `TryResolve<T>()`, `TryResolve(Type)`, and `TryResolve<T>(string keyName)` — the variants whose fallback runs after `GetService` returns null. The required `Resolve(...)` overloads are unaffected: their `?.` chain only reaches `provider` when there is no `HttpContext`.

## Tests

Added `ServiceResolver` regression tests (generic / non-generic / keyed `TryResolve`) that resolve an unregistered service with an active request scope while the captured root provider is disposed — they throw `ObjectDisposedException` without the fix and pass with it. All existing `ServiceResolver` tests continue to pass.

### DIFF
--- a/Src/Core/ServiceResolver/ServiceResolver.cs
+++ b/Src/Core/ServiceResolver/ServiceResolver.cs
@@ -64,17 +64,27 @@ sealed class ServiceResolver(IServiceProvider provider, IHttpContextAccessor? ct
         => ctxAccessor?.HttpContext?.RequestServices.GetRequiredService(typeOfService) ??
            provider.GetRequiredService(typeOfService);
 
+    //when a request scope is active, resolve solely from it instead of falling back to the captured root 'provider'.
+    //the request scope is a child of the root, so it already sees every registered service; the '?? provider' fallback
+    //would only ever return null for an unregistered service anyway. but 'provider' is a process-wide captured root, and
+    //in multi-host setups (e.g. several WebApplicationFactory instances in one test run) the static ServiceResolver.Instance
+    //can outlive the host that set it - querying that disposed root throws ObjectDisposedException. only fall back to the
+    //root when there is no active request scope (background command/event execution or unit-test setup).
+
     public TService? TryResolve<TService>() where TService : class
-        => ctxAccessor?.HttpContext?.RequestServices.GetService<TService>() ??
-           provider.GetService<TService>();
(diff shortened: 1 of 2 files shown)


dj-nitehawk:
thank you!

PR #1119, opened 2026-06-09
Added warmup logic for endpoints

SjorsGoering:
## Summary

- Adds a startup warmup phase that eagerly initializes expensive per-endpoint components (validators, mappers, request binders, compiled property setter delegates) and event buses so the first real requests don't pay the cold-start cost.
- Refactors `IEndpointFactory.Create` to accept `IServiceProvider` instead of `HttpContext`, decoupling instantiation from the HTTP pipeline and making it usable outside of a live request (e.g. during warmup).
- Adds `EndpointOptions.WarmupFilter` — an optional predicate to control which endpoints participate in warmup.

## Changes

- **`IEndpointFactory` / `EndpointFactory`** — `Create` now takes `IServiceProvider` instead of `HttpContext`. Call sites updated in `EndpointBootstrap`, `MainExtensions`, and `Testing.Factory`.
- **`MainExtensions.Warmup`** — new static method called after `MapFastEndpoints`; iterates `EndpointData.Found`, resolves binders, validators, mappers, object factories, and calls the endpoint factory once per definition.
- **`MessagingExtensions.WarmupMessaging`** — resolves each registered `EventBus<T>` at startup so handler registrations are JIT-compiled before traffic arrives.
- **`EndpointOptions.WarmupFilter`** — opt-in predicate; when set, only matching endpoints are warmed up.


dj-nitehawk:
made warmup opt-in and prevented endpoint instances from being created during warmup because the root service provider was being used for that. which is typically problematic for scoped service resolving. the alternative would be to always create a service scope when instantiating endpoints during warmup. but i think not creating endpoints at all is a better choice. thoughts?

SjorsGoering:
It looks like endpoint instances warmup  already happens in MapFastEndpoints since it calls epFactory.Create(def, httpCtx); in the endpoints.Found loop. So actually endpoints are already created/warmed-up at startup.

Looking at how the MapFastEndpoints works with the endpoints.Found loop, I think moving the opt-in warmup logic inside it makes more sense since:
- EndpointData does not need to be resolved again
- Cfg.EpOpts.Filter logic is not duplicated
- endpoints.Found loop is not duplicated

I've pushed the changes for this change, let me know what you think.
btw noticed a typo in a method (WarmUp -> Warmup)

dj-nitehawk:
that's a nice improvement.
the tests weren't passing, i fixed that.
also had to skip binder warmup for value type request dtos, because native aot DI doesn't support that.
give it a final look over...
i think it's ready to merge now...

SjorsGoering:
Good catch with the AOT code path.
I found some naming inconsistencies, that I changed. 
There's one thing that I'm not sure yet how to solve yet, which is executing the Messaging library warmup code via the main library. An approach could be to call the Messaging warmup logic when the main library warmup is requested.

Src/Library/Main/MainExtensions.cs
```diff
@@ -210,6 +210,9 @@ public static IEndpointRouteBuilder MapFastEndpoints(this IEndpointRouteBuilder
             }
         }
 
+        if (Cfg.EpOpts.WarmupRequested)
+            MessagingExtensions.WarmupMessaging(app.ServiceProvider);
```
dj-nitehawk, line 214: @SjorsGoering this should do it

SjorsGoering: Indeed it does

SjorsGoering:
I've also added the warmup logic to the JobQueue library registration side, in case someone only uses FastEndpoints.JobQueue and wants to use the warmup functionality.
Let me know what you think.

dj-nitehawk:
looks good now.
merging...
thank you!

PR #1121, opened 2026-06-14
Fix concurrent WAF host builds by serializing the host creation in Ap…

### LINKED ISSUE #1120: `AppFixture` builds the SUT host concurrently under parallel test classes → intermittent `System.Text.Json` corruption (`ArgumentException`/`NullReferenceException`) + silent `Services` vs `Server.Services` split
### Summary

When multiple xUnit test classes share the **same** `AppFixture<TProgram>` type and run in parallel (the
default), the SUT host is built **concurrently**. `WafCache` guarantees a single `WafWrapper` *object*, but the
host *build* happens later and is not synchronized: every fixture instance calls `_wafInstance.CreateClient()`,
and `WebApplicationFactory`'s first‑call build (`EnsureServer`/`StartServer`) lets N classes run
`CreateHost → Program.Main → UseFastEndpoints → MapFastEndpoints` at the same time. `MapFastEndpoints` mutates a
process‑shared `System.Text.Json` configuration list, so the concurrent builds corrupt it.

This contradicts the documented guarantee ("never boot up more than one SUT/WAF no matter how many test‑classes
are using an AppFixture").

### Environment

- `net10.0`
- `FastEndpoints` / `FastEndpoints.Testing` 8.1.0 (also observed on 7.1.1)
- `xunit.v3` 3.2.2

### Repro

Minimal repro repo: [OpenSource.FastEndpoints.Testing.TimeProvider](https://github.com/terryaney/OpenSource.FastEndpoints.Testing.TimeProvider)
- (`Tests/WafCacheRaceRepro.cs`): one shared `AppFixture<Program>` consumed by 25 independent `TestBase<App>`
classes, plus a counted singleton (`SutProbe`) to surface how many SUTs were actually built.

Powershell:
```powershell
& {
	cls
	$runs = 20
	$errors = 0
	for ($i = 1; $i -le $runs; $i++) {
		Write-Host ("Testing run {0}... " -f $i) -NoNewline
		$output = & dotnet test Tests -v minimal --nologo --no-restore 2>&1
		$exitCode = $LASTEXITCODE

		if ($exitCode -eq 0) {
			Write-Host "SUCCESS" -ForegroundColor Green
		} else {
			$errors++
			Write-Host "FAILED" -ForegroundColor Red
			Write-Host ("----- BEGIN FAILURE OUTPUT (run {0}) -----" -f $i) -ForegroundColor Yellow
			$output
			Write-Host ("----- END FAILURE OUTPUT (run {0}) -----" -f $i) -ForegroundColor Yellow
		}
	}

	$failureRate = if ($runs -gt 0) { [math]::Round(($errors * 100.0) / $runs, 1) } else { 0.0 }
	Write-Host ""
	Write-Host ("{0} of {1} test runs failed, {2}% failure rate" -f $errors, $runs, $failureRate) -ForegroundColor Cyan

	if ($errors -gt 0) { $global:LASTEXITCODE = 1 } else { $global:LASTEXITCODE = 0 }
}
```

Repro rate ≈ 23% (7/30) with 25 classes on an 8‑core box.

### Observed (all the same root cause)

```
System.ArgumentException : Source array was not long enough...
   at System.Collections.Generic.List`1.Insert
   at System.Text.Json.Serialization.ConfigurationList`1.Insert
   at FastEndpoints.MainExtensions.MapFastEndpoints
   at FastEndpoints.MainExtensions.UseFastEndpoints
   at Program.<Main>$
   at FastEndpoints.Testing.AppFixture`1.ConfigureAppHost / WafWrapper.CreateHost
   at Microsoft.AspNetCore.Mvc.Testing.WebApplicationFactory`1.StartServer / CreateClient
   at FastEndpoints.Testing.AppFixture`1.Xunit.IAsyncLifetime.InitializeAsync
```

Other intermittent signatures from the same run loop:
- `System.NullReferenceException` at `System.Text.Json…JsonTypeInfoResolverChain.GetTypeInfo` (request serialization)
- `System.NotSupportedException : JsonTypeInfo metadata for type '…' was not provided by TypeInfoResolver…`
- **Silent (no exception):** `AppFixture.Services` and `AppFixture.Server.Services` resolve **different**
  singleton instances from the same fixture — any test asserting on a resolved singleton (e.g. a fake
  `IEmailService`) then reads a different instance than the request pipeline used.

### Expected

One SUT host, built once, shared by all classes that use the fixture type. `AppFixture.Services` and
`AppFixture.Server.Services` should resolve the same singletons.

NOTE: I have independently confirmed this approach: pre‑warming the WAF once (single‑threaded, in an assembly
fixture) before the parallel classes start turned a ~32% failure rate (up to 4 distinct WAFs) into 0 failures
across 85 consecutive runs in a real DB‑backed suite.

terryaney:
Fixes #1120 

`AppFixture<TProgram>.InitializeAsync()` (in `AppFixture.Waf.cs`) does, for **every** fixture instance:

```csharp
_wafInstance = (WebApplicationFactory<TProgram>)await WafCache.GetOrAdd(tDerivedFixture), _ => new(WafInitializer));
Client = _wafInstance.CreateClient();   // <-- triggers the host BUILD, per instance
await SetupAsync();

async Task<object> WafInitializer()
{
    await PreSetupAsync();

    return new WafWrapper(ConfigureAppHost).WithWebHostBuilder(
        b =>
        {
            b.UseEnvironment("Testing");
            b.ConfigureTestServices(ConfigureServices);
            ConfigureApp(b);
        });
}
```

- `WafCache` + `AsyncLazy` correctly serialize creation of the **`WafWrapper` object** — that part is fine.
- But `WithWebHostBuilder(...)` does **not** build the host. The host is built lazily on first access.
- Every test class's fixture instance then calls `_wafInstance.CreateClient()` on the **same shared**
  `WafWrapper`. The deferred host build is **not serialized across these concurrent first callers**, so *N*
  parallel classes run `CreateHost` → `Program.Main` → `UseFastEndpoints` → `MapFastEndpoints` **at the same
  time**, racing on a shared `System.Text.Json` `ConfigurationList`.

So the cache prevents multiple `WafWrapper` objects, but not multiple concurrent host **builds**.

The fix:
```csharp
async Task<object> WafInitializer()
{
    await PreSetupAsync();

    var waf = new WafWrapper(ConfigureAppHost).WithWebHostBuilder(
        b =>
        {
            b.UseEnvironment("Testing");
            b.ConfigureTestServices(ConfigureServices);
            ConfigureApp(b);
        });

    // Build the single host HERE, while still inside the AsyncLazy, so it is serialized across all fixture
    // instances. Without this, each instance's `_wafInstance.CreateClient()` triggers the lazy host build, and
    // those builds are not serialized across parallel first callers — N test classes build the host
    // concurrently, racing on shared System.Text.Json config in MapFastEndpoints.
    _ = waf.Server;

    return waf;
}
```

### DIFF
--- a/Src/Library/changelog.md
+++ b/Src/Library/changelog.md
@@ -189,6 +189,12 @@ app.UseJobQueues(o => o.Warmup());
 
 ## Fixes 🪲
 
+<details><summary>'AppFixture' concurrent WebApplicationFactory host build issue</summary>
+
+`AppFixture` now builds the cached `WebApplicationFactory` host while the shared initializer is still serialized. This prevents parallel first-use test fixtures from each triggering their own lazy host build through `CreateClient()`, avoiding races during FastEndpoints startup configuration.
+
+</details>
+
 <details><summary>'ValidationSchemaProcessor' concurrency issue when generating multiple Swagger documents</summary>
 
 The `_childAdaptorValidators` instance field on the singleton `ValidationSchemaProcessor` was a plain non-thread-safe `Dictionary<string, IValidator>`. Concurrent Swagger document generation (e.g. multiple documents being built in parallel at startup) could cause race conditions on that shared dictionary. The field is now a local variable scoped to each `Process` call, eliminating shared mutable state entirely.
--- a/Src/Testing/AppFixture.Waf.cs
+++ b/Src/Testing/AppFixture.Waf.cs
@@ -195,10 +195,7 @@ public abstract partial class AppFixture<TProgram> : BaseFixture, IAsyncLifetime
                     ConfigureApp(b);
                 });
 
-            // Build the single host HERE, while still inside the AsyncLazy, so it is serialized across all fixture
-            // instances. Without this, each instance's `_wafInstance.CreateClient()` triggers the lazy host build, and
-            // those builds are not serialized across parallel first callers — N test classes build the host
-            // concurrently, racing on shared System.Text.Json config in MapFastEndpoints.
+            // build the single host inside the cached initializer to serialize parallel first callers.
             _ = waf.Services;
 
             return waf;
--- a/Src/Testing/AppFixture.Waf.cs
+++ b/Src/Testing/AppFixture.Waf.cs
@@ -199,7 +199,7 @@ public abstract partial class AppFixture<TProgram> : BaseFixture, IAsyncLifetime
             // instances. Without this, each instance's `_wafInstance.CreateClient()` triggers the lazy host build, and
             // those builds are not serialized across parallel first callers — N test classes build the host
             // concurrently, racing on shared System.Text.Json config in MapFastEndpoints.
-            _ = waf.Server;
+            _ = waf.Services;
 
             return waf;
         }
--- a/Src/Testing/AppFixture.Waf.cs
+++ b/Src/Testing/AppFixture.Waf.cs
@@ -187,13 +187,21 @@ public abstract partial class AppFixture<TProgram> : BaseFixture, IAsyncLifetime
         {
             await PreSetupAsync();
 
-            return new WafWrapper(ConfigureAppHost).WithWebHostBuilder(
+            var waf = new WafWrapper(ConfigureAppHost).WithWebHostBuilder(
                 b =>
                 {
                     b.UseEnvironment("Testing");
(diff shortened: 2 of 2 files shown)


dj-nitehawk:
merged...
thank you!

PR #1123, opened 2026-06-16
Add WriteOnlyWhenDifferent to SwaggerInitFile target

Plevi1337:
This PR contains a single change, it stops every build from updating the timestamp on the the SwaggerInitFile. The content of the file is not changing, it's hardcoded in the target, there is no point in always rewriting it. 
The timestamp update flagged the project that referenced the nuget as changed, causing unnecessary rebuilds on my solution. 

Sorry if something is missing, I've just found this minor inconvenience and wanted to upstream the fix, I've not found any contributing guidelines.


SjorsGoering:
I think a similar fix could be applied to FastEndpoints.OpenApi.targets

dj-nitehawk:
applied to FastEndpoints.OpenApi.targets as well.
thanks guys!

dj-nitehawk:
pushed `v8.2.0-beta.45` to nuget

PR #1127, opened 2026-06-16
Added cached compiled property getter in (recursive) DTO validation

SjorsGoering:
## Summary

- Adds a compiled expression-based `GetterForProp` to the reflection cache, mirroring the existing `SetterForProp` pattern
- Replaces the slow `PropertyInfo.GetValue` call in (recursive) DTO validation with the cached compiled getter
- Pre-compiles getters for top-level request DTO properties during endpoint warmup to avoid first-request JIT overhead

## Notes

Nested child objects encountered during recursive validation still get their getters compiled lazily on first use.
Pre-compiling them at warmup would require a recursive property-type traversal, which may not be worth the added complexity.


dj-nitehawk:
merged...
thank you very much!

PR #1132, opened 2026-06-28
Add refresh token service support for union-type returning endpoints

### LINKED ISSUE #597: Refresh Token with Union-Type Returning Handler
Hi,
in my application, I have a login endpoint using [Union-Type Returning Handler](https://fast-endpoints.com/docs/get-started#union-type-returning-handler), so my class is: 
```
sealed class Login : Endpoint<
    LoginRequest,
    Results<Ok<LoginResponse>, UnauthorizedHttpResult>> {

    ...

    public override async Task<Results<Ok<LoginResponse>, UnauthorizedHttpResult>> ExecuteAsync(LoginRequest req, CancellationToken ct)
    {
        ...
    }

    ...

}
```
I have an issue when I try to use [JWT Refresh Token Service](https://fast-endpoints.com/docs/security#jwt-refresh-tokens) to implement my refresh token endpoint and logic: the refresh endpoint response type can't be converted to login endpoint response type, because it's a union of typed result.

Is there a way to do something similar?
For the moment, I have to code both login and refresh endpoint independently, without using the token service.

EelcoLos:
## Problem

fixes #597. A login endpoint that returns a union-type result can't use the built-in JWT refresh-token service:

```csharp
sealed class Login : Endpoint<LoginRequest, Results<Ok<TokenResponse>, UnauthorizedHttpResult>>
```

`CreateTokenWith<TService>()` is constrained to the endpoint's own `TResponse`:

```csharp
protected Task<TResponse> CreateTokenWith<TService>(...)
    where TService : IRefreshTokenService<TResponse>
```

When `TResponse` is `Results<Ok<TokenResponse>, …>`, no `IRefreshTokenService` can satisfy that constraint (the union isn't a `TokenResponse`), so the endpoint simply can't call the helper. Today the only options are rolling your own token logic or dropping the union return type.

## Root cause

The refresh-token feature predates union-type results. `CreateTokenWith<TService>` ties the *token* response type to the *endpoint* response type through the single `TResponse` generic. That works for plain-DTO endpoints but is impossible for union-returning ones.

## Fix

Add a non-breaking overload that decouples the token response type from the endpoint's response type:

```csharp
protected Task<TTokenResponse> CreateTokenWith<TService, TTokenResponse>(
    string userId, Action<UserPrivileges> userPrivileges, object? request = null)
    where TService : IRefreshTokenService<TTokenResponse>
```

- Different generic arity, so no overload ambiguity and full source compatibility.
- The existing `CreateTokenWith<TService>` now delegates to it (no duplicated body).
- `request` is widened to `object?` because a union endpoint's request DTO is unrelated to `TokenRequest`; it flows only into the initial-creation hooks. The login path passes `isRenewal: false`, so the renewal-only `(TRequest)request` cast is never reached. A test locks that in.

Usage:

```csharp
public override async Task<Results<Ok<TokenResponse>, UnauthorizedHttpResult>> ExecuteAsync(LoginRequest req, CancellationToken ct)
{
    if (!await Valid(req)) return TypedResults.Unauthorized();
    var token = await CreateTokenWith<MyTokenService, TokenResponse>(req.UserId, p => p.Roles.Add("user"));
    return TypedResults.Ok(token);
}
```

## Tests

`TestHarness` fixture (`RefreshTokensTest/UnionLoginEndpoint.cs`) plus integration tests in `SecurityTests.cs`:

- `UnionLoginEndpointGeneratesCorrectToken`. The `#597` scenario: the `Ok<TokenResponse>` branch issues a valid JWT with the expected roles/permissions/claims, and asserts the renewal-only claim is **absent** (proving the login path doesn't trigger the renewal hook).
- `UnionLoginEndpointReturnsUnauthorizedBranch`. The union's `UnauthorizedHttpResult` branch still returns 401.

Changelog updated under **Improvements 🚀**.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,6 +1,6 @@
 [*]
 charset = utf-8
-end_of_line = crlf
+end_of_line = lf
 insert_final_newline = false
 indent_style = space
 indent_size = 4
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,4 @@
+* text=auto eol=lf
 *.verified.txt text eol=lf working-tree-encoding=UTF-8
 *.verified.xml text eol=lf working-tree-encoding=UTF-8
 *.verified.json text eol=lf working-tree-encoding=UTF-8
--- a/Src/Library/Endpoint/Endpoint.cs
+++ b/Src/Library/Endpoint/Endpoint.cs
@@ -315,7 +315,21 @@ public abstract partial class Endpoint<TRequest, TResponse> : BaseEndpoint, IEve
     /// <param name="userPrivileges">the user privileges to be embedded in the jwt such as roles/claims/permissions</param>
     protected Task<TResponse> CreateTokenWith<TService>(string userId, Action<UserPrivileges> userPrivileges, TRequest? request = default)
         where TService : IRefreshTokenService<TResponse>
-        => ((IRefreshTokenService<TResponse>)
+        => CreateTokenWith<TService, TResponse>(userId, userPrivileges, request);
+
+    /// <summary>
+    /// create the access/refresh token pair response with a given refresh-token service, decoupled from the endpoint's response type.
+    /// use this overload when the endpoint returns a union-type result (e.g. <c>Results&lt;Ok&lt;TokenResponse&gt;, UnauthorizedHttpResult&gt;</c>)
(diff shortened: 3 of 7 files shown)


dj-nitehawk:
looks good... merged...
thank you!

PR #1137, opened 2026-07-06
FrozenDictionary usages for build-once, read-many lookup caches

SjorsGoering:
## Use FrozenDictionary for build-once, read-many lookup caches

Converts several internal `Dictionary` fields to `System.Collections.Frozen.FrozenDictionary` across the request-binding, error-handling, and OpenAPI/Swagger doc-generation paths. Each converted table is built exactly once (per type, per assembly, or as a static field) and then read repeatedly — the shape `FrozenDictionary` is designed to speed up, at the cost of a slower one-time build.

### Changed

- **`RequestBinder<TRequest>`** — `_primaryProps`/`_formFileCollectionProps`, built once per DTO type, read on every request bound to that DTO.
- **`ErrorOptions.ProblemDetailsConfig`** — `_codeMap`, read on every ProblemDetails error response.
- **`ConstructorDefaultExtensions`** — per-`Type` constructor-default cache, read on every OpenAPI doc generation pass.
- **`XmlDocLookup`** — per-`Assembly` XML doc cache, read on every OpenAPI doc generation pass.
- **`OperationTransformer.Responses`** / **`Swagger.OperationProcessor`** — `_defaultDescriptions` status-code maps.
- **`AccessControlGenerator`** — the generated `Allow` class's `_permNames`/`_permCodes` permission lookup tables, read on every permission check in every consuming app. Required reshaping the generated `InitPermissions` partial method to populate local mutable dictionaries that get frozen and assigned in the static constructor (readonly fields can only be assigned from the declaring type's own constructor, and `InitPermissions` may have zero implementations in projects with no registered permissions).
- **`Swagger.GlobalConfig.RouteConstraintMap`** — read on every OpenAPI/Swagger doc generation pass to resolve route-constrained parameter types.

### Why

All of these dictionaries are populated once (at type load, in a static constructor, or cached once per key) and never mutated afterward, while reads happen on a hot or semi-hot path — request binding, error responses, or doc-generation passes that can run on every request depending on which OpenAPI pipeline is in use. That's exactly the build-once/read-many pattern `FrozenDictionary` trades slower construction for faster lookups on.

### Testing

- Reviewed each conversion site's build/read frequency to confirm none rebuild the dictionary more often than before, including verifying that Microsoft.AspNetCore.OpenApi rebuilds its document per request while NSwag caches it — both amortize the one-time build cost across many reads either way.
- `AccessControlGeneratorTests` (which compile the generator's actual emitted output) pass, including the zero-permission case where `InitPermissions` has no implementation.
- No behavior change intended anywhere — only the backing collection type for already-immutable-after-construction data.

### DIFF
--- a/Src/Generator/AccessControlGenerator.cs
+++ b/Src/Generator/AccessControlGenerator.cs
@@ -256,7 +256,7 @@ public class AccessControlGenerator : IIncrementalGenerator
             sb.w(
                 """
 
-                    static partial void InitPermissions()
+                    static partial void InitPermissions(Dictionary<string, string> permNames, Dictionary<string, string> permCodes)
                     {
 
                 """);
@@ -265,8 +265,8 @@ public class AccessControlGenerator : IIncrementalGenerator
             {
                 sb.w(
                     $"""
-                             _permNames[{CsString(p.Name)}] = {p.CodeExpression};
-                             _permCodes[{p.CodeExpression}] = {CsString(p.Name)};
+                             permNames[{CsString(p.Name)}] = {p.CodeExpression};
+                             permCodes[{p.CodeExpression}] = {CsString(p.Name)};
 
                      """);
             }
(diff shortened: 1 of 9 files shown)


dj-nitehawk:
two sites had to be reverted because they were public mutable dictionaries.
other than that, all good...
thank you!

PR #1138, opened 2026-07-08
Use FrozenSet lookups for permission/scope/claim-type checks

SjorsGoering:
## Summary

- Cache `AllowedPermissions`/`AllowedScopes`/`AllowedClaimTypes` into `FrozenSet`s in `MainExtensions.cs` before use in the `RequireAssertion` claim-matching delegates, replacing per-check `Enumerable.Contains(..., comparer)` calls with O(1) frozen-set lookups. Continues the frozen-collection perf work from #1137.
- Preserves existing comparison semantics: permissions stay ordinal (case-sensitive), scopes and claim types stay ordinal-ignore-case.
- Adds `ComparerSemanticsTests` test endpoints and `SecurityTests` cases to lock in that casing behavior (permission mismatch → 403, scope/claim-type casing mismatch → still allowed).
- Updates OpenAPI snapshot fixtures (`release-0/1/2.json`) for the new test endpoints.

### DIFF
--- a/Src/Library/Main/MainExtensions.cs
+++ b/Src/Library/Main/MainExtensions.cs
@@ -1,4 +1,5 @@
 using System.Collections.Concurrent;
+using System.Collections.Frozen;
 using System.Diagnostics.CodeAnalysis;
 using System.Text;
 using System.Text.Json.Serialization;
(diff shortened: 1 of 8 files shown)


dj-nitehawk:
merged, thank you very much!

PR #1139, opened 2026-07-09
Optimize RequestBinder claim/permission binding logic

SjorsGoering:
## Summary

Follow-up perf work to #1137 (frozen auth-policy lookups). Optimizes `RequestBinder<TRequest>`'s claim and permission binding to avoid building a fresh heap-allocated collection from *all* of the user principal's claims on every request.

## Problem

- `BindUserClaims` built a per-request `Dictionary<string, List<string>>` populated from every claim on the principal, even claim types the request DTO never binds via `[FromClaim]`.
- `BindHasPermissionProps` built a per-request `HashSet<string>` from every permission claim on the principal, even when the DTO only checks a handful of specific permissions via `[HasPermission]`.

Both scale with the size of the user's claims/permissions rather than the size of the DTO, which is wasteful for principals carrying many claims (common with JWTs) or large permission sets (common in RBAC setups).

## Change

- Added two static `FrozenDictionary<string, int[]>` indices, built once per DTO type in `RequestBinder`'s static constructor:
  - `_claimTypeToPropIndices` — claim type (`OrdinalIgnoreCase`) → indices into `_fromClaimProps`
  - `_permissionToPropIndices` — permission value (`Ordinal`, matching existing pinned comparer semantics) → indices into `_hasPermissionProps`
  - Both support multiple properties sharing the same identifier (e.g. two `[FromClaim("dept")]` properties), preserving prior behavior.
- `BindUserClaims` now does a single pass over the principal's claims, doing an O(1) frozen-dictionary lookup per claim and only lazily allocating a `List<string>` for properties that actually match — no dictionary entries for irrelevant claim types.
- `BindHasPermissionProps` now uses a `stackalloc bool[]` matched-flags buffer (falls back to a heap array only if the DTO declares more than 128 `[HasPermission]` properties) instead of a `HashSet<string>` — zero heap allocation in the common case.

## Testing

Added `[FromClaim]` unit tests to `RequestBinderTests.cs`:
- single claim value bound to a property
- multiple claim values bound to a collection property
- two properties bound from the same claim type (regression test for the `int[]`-per-identifier index design)
- required claim missing → validation failure

Existing unit and integration suites pass unchanged.

## OKF

No update needed — internal-only change, no public API or observable binding-behavior change; the new indices follow the same build-once-at-static-init lifecycle as the existing `_primaryProps` cache, so no new caveat is introduced.

### DIFF
--- a/Src/Library/Binder/RequestBinder.cs
+++ b/Src/Library/Binder/RequestBinder.cs
@@ -32,8 +32,12 @@ public class RequestBinder<TRequest> : IRequestBinder<TRequest> where TRequest :
     static readonly List<SecondaryPropCacheEntry> _fromHeaderProps = [];
     static readonly List<SecondaryPropCacheEntry> _fromCookieProps = [];
     static readonly List<SecondaryPropCacheEntry> _hasPermissionProps = [];
-    static FrozenDictionary<string, int[]> _claimTypeToPropIndices = FrozenDictionary<string, int[]>.Empty; //key: claim type (OrdinalIgnoreCase); value: indices into _fromClaimProps
-    static FrozenDictionary<string, int[]> _permissionToPropIndices = FrozenDictionary<string, int[]>.Empty; //key: permission value (Ordinal); value: indices into _hasPermissionProps
+
+    //key: claim type (OrdinalIgnoreCase); value: indices into _fromClaimProps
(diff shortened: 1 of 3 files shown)


dj-nitehawk:
looks good...
thank you very much!

PR #1140, opened 2026-07-11
Optimize response serialization hotpath

SjorsGoering:
## Summary

Optimizes the response-serialization hot path to cut per-response work in two spots:

- **`SerializerOptions.cs`** — the charset-suffixed content-type string (e.g. `"application/json; charset=utf-8"`) was being rebuilt via string interpolation on every JSON response. It's now cached for the two most common content types (`application/json`, `application/problem+json`) and recomputed only when `CharacterEncoding` is configured, instead of on every request.
- **`HttpResponseExtensions.cs`** — `ApplyGlobalResponseModifier` changed from `async Task` to a plain `Task`-returning method (`GlobalResponseModifierAsync?.Invoke(...) ?? Task.CompletedTask`), avoiding an async state machine allocation on the common no-modifier-configured path across all ~17 call sites.

## Testing

Existing unit and integration suites pass unchanged (no behavior change, response bytes/headers identical).

## OKF

No update needed — internal-only change, no public API or observable behavior change.


dj-nitehawk:
thank you!

PR #1141, opened 2026-07-11
Regex/hash cleanups: source-generated regexes, LINQ-free ACL hash, safer PropNames

SjorsGoering:
## Summary

Four independent, low-risk cleanups — no public API or observable behavior change:

- **`Src/Swagger/OperationProcessor.cs`** — `TagName`'s inline `Regex.Replace(val, "[^a-zA-Z0-9]", "")` converted to a `[GeneratedRegex]` partial method (`NonAlphaNumericRegex`), matching the other source-generated regexes already in this file.
- **`Src/Generator.Cli/Extensions.cs`** — dropped the redundant `RegexOptions.Compiled` flag from the already-`[GeneratedRegex]`-attributed `SanitizationRegex()`. A source-generated regex already gets compiled-equivalent codegen at build time, so the flag was a pure no-op runtime cost.
- **`Src/Library/Extensions/ReflectionExtensions.cs`** — `PropNames<T>` replaced `a.ToString().Split('.')[1]` with `m.Member.Name`. The `MemberExpression` was already pattern-matched two lines above, so stringifying-then-splitting the whole expression was wasted work.
- **`Src/Library/Endpoint/Endpoint.Base.cs`** and **`Src/Generator/AccessControlGenerator.cs`** — both copies of `GetAclHash` (which must stay algorithmically identical per their own "if modifying this algo, update ... also" comments) had their `.Where(char.IsLetterOrDigit).Take(3).Select(char.ToUpper).ToArray()` LINQ chain replaced with an equivalent manual loop. Same predicate, same short-output behavior when fewer than 3 alphanumeric characters exist. The runtime (`Endpoint.Base.cs`) version uses a `stackalloc char[3]` buffer; the generator's (`AccessControlGenerator.cs`, netstandard2.0, build-time only) version uses a plain `char[3]` array to avoid adding a `Span<T>`/`System.Memory` assumption to an analyzer package.

## Testing

- Added `Tests/UnitTests/FastEndpoints/ReflectionExtensionsTests.cs` — direct unit tests for `PropNames<T>`.
- Full unit suite (264/264), integration suite (199/199), and `AccessControlGeneratorTests` (8/8, cross-checks the generator's computed codes) pass.

## OKF

No update needed — internal cleanups only, no public API or observable behavior change.

### DIFF
--- a/Src/Generator.Cli/Extensions.cs
+++ b/Src/Generator.Cli/Extensions.cs
@@ -23,7 +23,7 @@ static partial class Extensions
         }
     }
 
-    [GeneratedRegex("[^a-zA-Z0-9]+", RegexOptions.Compiled)]
+    [GeneratedRegex("[^a-zA-Z0-9]+")]
     private static partial Regex SanitizationRegex();
 
     internal static string Sanitize(this string input, string replacement = "_")
--- a/Src/Generator/AccessControlGenerator.cs
+++ b/Src/Generator/AccessControlGenerator.cs
@@ -593,7 +593,16 @@ public class AccessControlGenerator : IIncrementalGenerator
             using var sha256 = SHA256.Create();
             var base64Hash = Convert.ToBase64String(sha256.ComputeHash(Encoding.UTF8.GetBytes(input.ToUpperInvariant())));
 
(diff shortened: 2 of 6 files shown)


dj-nitehawk:
thanks!

PR #1143, opened 2026-07-14
Fix collection-element-type regression breaking byte-array binding

SjorsGoering:
## Summary

Fixes a binding regression introduced by the `GetCollectionElementType()` rewrite in the nested-DTO-warmup work: query/form properties typed as a **concrete, non-generic class deriving from a generic byte collection** (e.g. `class ByteBuffer : List<byte> {}`) were being routed to base64 decoding (`TryParseByteArray`) instead of JSON/CSV array parsing (`TryParseCollection`), breaking previously-working binding for that shape.

## Root cause

`BinderExtensions.cs`'s value-parser compilation decided "route to base64" based on `type.GetCollectionElementType() == Types.Byte` — i.e. "does this collection's element type happen to be byte?" But `TryParseByteArray` returns a raw `byte[]`, which can only be assigned to a property that is *actually* `byte[]` — not to `List<byte>`, `ByteBuffer`, or anything else merely shaped like "a collection of bytes." Once `GetCollectionElementType()` learned to resolve element types via interface-walking (to correctly support derived/custom collection types elsewhere), it started reporting `byte` for these types too, silently flipping their parser routing and either failing to parse (invalid base64) or throwing `InvalidCastException` when the setter tried to assign a `byte[]` into a non-`byte[]` property.

## Fix

- `Src/Library/Binder/BinderExtensions.cs` — restored the precise precondition: `type.IsArray && type.GetElementType() == Types.Byte`. Only a genuine `byte[]` array routes through the base64 path now; every other byte-elemented collection type goes back through `TryParseCollection` (JSON/CSV array), matching pre-regression behavior.
- `Src/Library/Testing/HttpClientExtensions.cs` — deduplicated: `HasCompatibleShape`'s local `GetCollectionElementType` helper was an exact copy of the shared `Type.GetCollectionElementType()` extension; removed it in favor of calling the shared method directly.

## Testing

Added two `RequestBinderTests.cs` cases:
- `ByteCollectionBindsAsBase64NotJsonArray` — a genuine `byte[]` property still binds correctly from a base64-encoded query value.
- `DerivedByteCollectionBindsAsJsonArrayNotBase64` — a `ByteBuffer : List<byte>` property binds correctly from a JSON-array-style query value (`[1,2,3]`), guarding against the regression this PR fixes.

Full unit suite passes (270/270).

## OKF

No update needed — internal binder fix, no public API or observable-behavior change beyond restoring prior binding behavior for this DTO shape.

### DIFF
--- a/Src/Library/Binder/BinderExtensions.cs
+++ b/Src/Library/Binder/BinderExtensions.cs
@@ -147,7 +147,7 @@ static class BinderExtensions
                     var interfaces = type.GetInterfaces();
 
                     return interfaces.Contains(Types.IEnumerable) && !interfaces.Contains(Types.IDictionary) //dictionaries should be deserialized as json objects
-                               ? type.GetCollectionElementType() == Types.Byte
+                               ? type.IsArray && type.GetElementType() == Types.Byte
                                      ? input => new(TryParseByteArray(input, out var res), res)
                                      : input => new(TryParseCollection(input, type, out var res), res)
                                : input => new(TryParseObject(input, type, out var res), res);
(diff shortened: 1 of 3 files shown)


dj-nitehawk:
merged... thank you!

PR #1145, opened 2026-07-16
Export OpenAPI documents as '.http' files

### LINKED ISSUE #992: Feature Request: Built-in HttpGenerator for FastEndpoints (generate .http files from endpoints/Swagger)
I’d like to propose adding built-in support for generating .http files (JetBrains Rider / VS Code REST Client format) directly from FastEndpoints + FastEndpoints.Swagger.
The idea is to provide developers an easy way to auto-generate request collections for testing APIs, similar to how Swagger UI or Postman collections can be generated.

**Motivation**
- Many .NET developers use Rider or VS Code with .http files for local testing.
- Currently, Swagger/OpenAPI can be exported, but generating .http files requires external tooling or custom scripts.
- A first-class integration in FastEndpoints would simplify workflow: every endpoint could carry its own HttpGen metadata and the library would automatically produce .http files.
- 

Proposal

Introduce a new package, e.g. FastEndpoints.HttpGen, with:

Endpoint metadata extensions (used in Configure()):

```
Definition
    .HttpGenFile("chapters")
    .HttpGenAuthProfile("keycloak")
    .HttpGenName("CreateChapter")
    .HttpGenSaveVar("/id", "chapterId");

```

OperationProcessor integration with FastEndpoints.Swagger:

- Reads endpoint metadata
- Writes x-httpgen-* vendor extensions into OpenAPI spec

Runtime generator (middleware or CLI):

- Reads Swagger doc
- Generates .http files into Requests/*.http
- Groups by file (e.g. per tag, per auth profile, or explicit HttpGenFile metadata)

Single-line registration in Program.cs:
```
builder.Services.AddFastEndpointsWithHttpGen(o =>
{
    o.DocumentName = "v1";
    o.OutputDir = "Requests";
    o.GenerateRoutePath = "/__gen/http";
});
```
**Benefits**

- Developer convenience: quick local testing without Postman.
- Consistency: requests are generated from the same OpenAPI as Swagger.
- Flexibility: metadata per-endpoint allows different auth profiles, grouping, saving response values as variables.
- Minimal setup: one-line configuration, everything else automatic.

EelcoLos:
Closes #992 (scoped down, see [comment](https://github.com/FastEndpoints/FastEndpoints/issues/992#issuecomment-4989321037)).

## Summary

Adds a sibling `.http` export to `FastEndpoints.OpenApi`'s existing `.json` export mechanism (`ExportOpenApiDocsAndExitAsync` / `--export-openapi-docs`). Same in-memory `OpenApiDocument`, same export-and-exit shape, same MSBuild AOT-publish hook. No new package, no external CLI dependency.

```csharp
await app.ExportHttpFilesAndExitAsync("v1");
```
```
dotnet run --export-http-files true
```

## Tasks

- [x] Comment posted on #992
- [x] `HttpFileExporter.cs` + schema-walker body placeholders
- [x] `OpenApiExporter.cs` refactor + `ExportHttpFilesAndExitAsync`
- [x] `Extensions.cs` flag/entry point
- [x] `.targets` MSBuild target
- [x] Integration tests + golden files
- [x] Changelog entry
- [x] FE-Docs page update ([FastEndpoints/Documentation#38](https://github.com/FastEndpoints/Documentation/pull/38))
- [x] Reviews (`/review`, `/code-review-perfect`, `/ponytail-review`, `/security-review`) addressed

## Format

One `.http` file per OpenAPI document, written alongside the `.json` export. REST Client format: path/query/header params as `{{var}}` placeholders, request body as a property-keyed JSON skeleton walked from the schema (falls back to `{}` only for genuinely composed/polymorphic schemas — the common nullable-`$ref` idiom is unwrapped rather than treated as polymorphism).

## Review notes

Ran a 4-way review pass (functional/spec, PERFECT framework, over-engineering, security) and fixed what it found:
- Security: no findings — same trust model as the existing JSON exporter/MSBuild target it mirrors.
- Correctness: the schema-walker's cycle guard was marking every visited schema instead of just the current recursion path, so a DTO referenced twice non-cyclically (e.g. the same `Address` type used in two properties) got nulled out the second time — fixed to be path-scoped.
- Spec: the common nullable-`$ref` idiom (`oneOf: [null, ref]`) was being misclassified as real polymorphism and falling back to `{}` far more often than intended — now unwrapped to derive real placeholder fields.
- Over-engineering: deduped the JSON/`.http` export try/catch/write/log shape into one shared helper.


dj-nitehawk:
merged after some modifications. thank you very much 🙏

PR #1146, opened 2026-07-23
Cache bindable props per DTO type in binder hot path

SjorsGoering:
## Summary

Caches the materialized set of bindable properties per request DTO type so the binder hot path stops allocating (and locking) on every call.

`BinderExtensions.BindableProps()` previously returned `ConcurrentDictionary.Keys`. Accessing that property does two costly things on every call: it acquires **all** of the dictionary's internal locks and then copies the keys into a freshly allocated `List<PropertyInfo>` snapshot. This method is called on the request-binding hot path (complex binders + data-annotation validation recursion), so both the locking and the allocations add up under load.

This PR materializes the keys into a `PropertyInfo[]` once and caches it on `TypeDefinition.BindableProps`, reusing it on subsequent calls — no repeated locking, no repeated allocation.

## Changes

- **`ReflectionCache.cs`** — added `internal PropertyInfo[]? BindableProps` to `TypeDefinition` to hold the cached snapshot.
- **`BinderExtensions.cs`** — `BindableProps()` now returns `PropertyInfo[]` and lazily populates/reuses the cached array (`typeDef.BindableProps ??= ...Keys.ToArray()`).
- **`RequestBinder.cs`** — updated call site from `.Count` to `.Length`.
- **`HttpClientExtensions.cs`** — updated call site from `.Count` to `.Length`.

## Behavior & assumptions

The cache assumes no properties are added to `TypeDefinition.Properties` after startup, which the public API does not do. The materialized array is treated as immutable for the lifetime of the type definition. Note the `.Keys` snapshot is still taken exactly once (on first access) to populate the cache — the locking cost is paid a single time rather than on every bind.

## Risk

Low. Purely a locking/allocation optimization on an existing code path; the returned property set is unchanged, and both call sites were updated to the array-typed API.

### DIFF
--- a/.okf/architecture.md
+++ b/.okf/architecture.md
@@ -51,6 +51,7 @@ Attributes / Messaging.Core
 - **Security/OpenApi/OData/AspVersioning** reference Library (addons on top of core HTTP).
 - **Generator** references Attributes only (analyzer package); consumers reference Generator as analyzer.
 - **Agents** (`Mcp`, `A2A`) reference Library; share internal types via linked `Src/Agents/Shared/*.cs` (not a separate NuGet).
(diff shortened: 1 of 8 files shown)


dj-nitehawk:
merged...
thank you!

PR #1147, opened 2026-07-31
Reduce per-request closure allocations

SjorsGoering:
## Summary

Removes a heap allocation that happened on **every request**, including successful ones.

`ValidationFailed` was an `async` local function inside `Endpoint<TRequest, TResponse>.ExecAsync` that captured the enclosing locals `req`, `ranPreProcessors` and `ct`. When an async local function captures an enclosing local, Roslyn has to emit a heap `<>c__DisplayClass` to hold those variables, and that closure is allocated when the captured locals come into scope (method entry), not when the local function is invoked. So every request paid for it, even though the helper only ever runs on the two failure paths (`JsonException` and `ValidationFailureException`).

The helper is now a private method that takes those three values as parameters, so nothing is captured except `this`, which is free (it compiles down to a plain instance method).

## Changes

- `Src/Library/Endpoint/Endpoint.cs`: the `ValidationFailed` local function becomes a private method `HandleValidationFailure(Exception x, int statusCode, TRequest req, bool ranPreProcessors, CancellationToken ct)`, placed directly after `ExecAsync` and called from the two `catch` blocks.
- `Src/Library/changelog.md`: entry under Improvements.
- `.okf/gotchas.md`: note so this does not get folded back into a capturing local function.

The method could not keep the name `ValidationFailed`, because that is already a `public bool` property of the same partial class (`Endpoint.Validation.cs`), which makes a same-named private method a CS0102 duplicate. The local function got away with the name only because it lived in its own scope.

Making it a separate method rather than a local function with parameters was deliberate: both fix the allocation identically, but with a separate method the enclosing locals are simply out of scope, so the closure cannot be reintroduced by tidying the parameter list away. It also lets the parameters keep their natural names (`req`, `ranPreProcessors`, `ct`) instead of renaming them to dodge CS0136 shadowing.

## Behavior

Unchanged. The two call sites pass the values immediately inside the `catch` blocks, and the helper's first read of them happens synchronously before any `await`, so there is no path on which a captured variable could have differed from the passed argument. Pre-processor skipping, `DoNotCatchExceptions` rethrow, and the `ResponseStarted` guard all behave exactly as before.

## Measurements

Compiled metadata of `Src/Library/bin/Release/net10.0/FastEndpoints.dll`, before and after:

```
before:  Endpoint`2 nested type  <>c__DisplayClass0_0   CLASS (heap)   [<>4__this, ct, ranPreProcessors, req]
after :  no closure type; <ExecAsync>d__0 (struct) holds <req>5__2 / <ranPreProcessors>5__3 directly
```

Allocation per call, measured with `GC.GetTotalAllocatedBytes(precise: true)` over 500k iterations on a standalone repro of the same `try`/`catch`/`catch`/`finally` plus `await` shape (workstation GC):

| Config | Before | After | Saved |
|---|---|---|---|
| success, awaits complete synchronously | 120 B | 72 B | 48 B |
| success, real suspension (state machine boxed) | 375 B | 306 B | 48 B+ |
| validation failure, awaits complete synchronously | 320 B | 272 B | 48 B |
| validation failure, real suspension | 576 B | 528 B | 48 B |

48 bytes is the object header (16) plus `this` (8), `req` (8), `ct` (8) and the `bool` (1), padded.


dj-nitehawk:
merged... thank you!

PR #1151, opened 2026-08-04
Extend warmup + fix indexer handling in BindableProps()

SjorsGoering:
## Summary

`ComplexSourceBinder`'s per-property binding metadata (field names, type-kind flags, setters, value parsers, `List<T>` factories) is cached lazily on the reflection cache, but nothing populated it ahead of time — `EndpointWarmup` never touched it, so the first request to any `[FromForm]`/`[FromQuery]` endpoint still paid the full reflection cost.

This adds a warmup pass that preloads it. Building that pass surfaced a pre-existing crash — indexer properties leaking into `BindableProps()` — which is fixed here as well, since it's the reason the traversal couldn't be written safely.

## 1. Warmup preloads the complex-bind cache

`EndpointWarmup.PrecompileComplexBindGraph` walks the object graph rooted at each `[FromForm]`/`[FromQuery]` property and populates everything the binder reads per request.

It's kept separate from the existing `PrecompileValidatableType` pass, which deliberately avoids setters and nested `ObjectFactory` because those fail for abstract or non-constructible declared types that runtime validation still supports.

Roots are discovered off the DTO directly, not off `RequestBinder<T>`'s `_fromFormProp`/`_fromQueryProp` — a user-supplied `IRequestBinder<>` registration means that static ctor never runs during warmup, so those caches would be null.

Coverage against every cache lookup on the bind path:

| Per-request work | Preloaded |
|---|---|
| `propCache.PropType.ObjectFactory()` | ✅ |
| `tParent.BindableProps()` | ✅ |
| `tParent.ComplexBindMeta(prop)` — every prop, every node | ✅ |
| `meta.UnderlyingType.ObjectFactory()` — nested complex | ✅ |
| `tElement.ObjectFactory()` — complex collection elements | ✅ |
| `new FormFileCollection()` | n/a — plain allocation, no cache |

### Benefits

- **First request no longer pays reflection cost.** Attribute lookup (`FieldName()`), type classification, `SetterForProp` and `ValueParser` expression compilation, and `MakeGenericType` + `ObjectFactory` for `List<T>` all move to startup — for every node in the nested graph, not just the root.
- **No first-request lock contention.** `ComplexBindMeta` takes `lock (propDef)` for its initial multi-field fill. With `FieldName` (the publication sentinel) already set, readers hit the fast path and concurrent startup traffic no longer serialises per property.
- **Consistent with what warmup already promises.** Warmup already primes binders, validators, mappers, and validation getters; the complex-bind cache was the remaining gap.

Two shapes are skipped on purpose, because warmup must never convert a lazy, data-dependent failure into a startup failure:

- **Form-file properties** — metadata is published, but no recursion and no factory. `IFormFile` is never instantiated by the binder.
- **Non-instantiable nodes** (abstract, or no public declared-only ctor) — prop metadata is published, but no `ObjectFactory`. The binder only news a nested node up when `HasDataForPrefix` is true, so compiling those eagerly would turn a possibly-never-hit path into a boot crash. `IsInstantiable` mirrors the rule inside `ObjectFactory()`'s compile step, and the coupling is documented on the method so the two don't silently drift.

## 2. Fix: indexers are not bindable properties

`BindableProps()` matched indexers, because `List<T>.Item[int]` and any `public T this[int]` have public get *and* set accessors. `SetterForProp`/`GetterForProp` then compile `Expression.Property(obj, "Item")` with no index arguments, which throws:

```
ArgumentException: Incorrect number of arguments supplied for call to method 'T get_Item(Int32)'
```

This was never limited to the complex binder — **any request DTO declaring a public indexer failed at endpoint registration:**

| Path | Before | After |
|---|---|---|
| `BindableProps()` on a type with an indexer | `[Item, Label]` | `[Label]` |
| `RequestBinder<T>` static ctor | **throws** → registration fails | binds |
| `GetterForProp` (DataAnnotations recursion) | **throws** | works |
| `ComplexSourceBinder`, node with an indexer | **throws** at bind time | binds correctly |
| `ComplexSourceBinder`, `List<List<T>>` | **throws** `ArgumentException` | descriptive `NotSupportedException` |

Fixed centrally in `BinderExtensions.GetProperties` with `p.GetIndexParameters().Length == 0`. An indexer has no bindable field name and no compilable accessor, so it was never a candidate. This matches three other reflection sites already in the codebase — `ProtobufMarshaller.cs:130`, `OpenApiSchemaExtensions.cs:231`, `NuGetPackageTypeLoader.cs:305` — `BindableProps()` was the outlier.

The same filter (`IsIndexer: false`) is added to `ReflectionGenerator`, since the source generator pre-populates the identical cache and would otherwise bypass the runtime filter entirely.

**Benefit:** fixing this at the source repairs every consumer at once — the top-level binder, the complex binder, DataAnnotations validation recursion, warmup, and the Agents/MCP schema and Testing helpers that also enumerate `BindableProps()`.

### Nested collections now throw instead of silently binding an empty list

The indexer fix on its own turned `List<List<T>>` from a crash into a **silently empty list** — trading a loud bug for silent data loss. `BindCollectionType` now rejects the shape explicitly, matching the `NotSupportedException` it already throws for `T[]`:

```
Collection elements of type 'List`1' are not supported for complex form binding,
because they are themselves collections! Offender: [MatrixDto.Matrix]
```

The shape is genuinely unsupported: `BindComplexCollection` binds each element as an object with named properties and has no branch that recurses back into collection binding, so there is no key convention for `Matrix[0][0].Qty`.

Gated on a new cached `ElementIsCollection` flag combined with `ElementIsComplex`, so only the `BindComplexCollection` path is affected. `List<string>` and other simple-parsed element types are untouched.

**Also affected: `List<byte[]>` and `List<Dictionary<,>>`,** which previously returned an empty list rather than throwing. Verified empirically that neither ever bound data — `byte[]` has zero bindable properties, so the recursion terminated on its first iteration and produced an empty list every time. No working code can depend on them. It is still a non-throwing path becoming a throwing one, so it's flagged explicitly; the condition can be narrowed if those should stay silent.

### Not a behavior change

- Warmup stays opt-in via `c.Endpoints.Warmup()` and respects `WarmupFilter`. Apps that don't call it see no effect from the warmup change.
- The `lock (propDef)` in `ComplexBindMeta` is retained — it's still needed on the no-warmup path.
- Removing indexers from `BindableProps()` cannot break working code, because every consumer threw on reaching one.

## Testing

- **`ComplexSourceBinderTests.cs`** (new, 7 tests): indexers excluded from `BindableProps`; indexer-declaring nodes bind at root and nested; `RequestBinder<T>` construction succeeds; nested collections throw for both form and query sources; and an ordinary graph (`Name`, `Tags[n]`, `Nested.City`, `Items[n].Qty`) still binds correctly as a regression guard.
- **`WarmupTests.cs`** (+5 tests): form graph, query graph, abstract and form-file nodes staying uninstantiated, self-referencing graphs, and indexers/nested collections not failing startup. Includes `AssertGraphFullyPublished`, which walks the graph exactly the way `ComplexSourceBinder` does and asserts every property it would touch already has `FieldName`, `Setter`, and a `ValueParser`/`ListFactory` where applicable — so the tests pin the actual cache state, not a proxy. Each new test was confirmed red before its corresponding change.
- `InternalsVisibleTo("Unit.FastEndpoints")` added to `FastEndpoints.Attributes`, needed because the fields this work is about (`FieldName`, `ListFactory`, `ValueParser`) are internal to that assembly.
- Verified against the real harness by temporarily enabling `c.Endpoints.Warmup()` in `TestHarness/Web`: all 12 `[FromForm]`/`[FromQuery]` roots warm at startup without throwing, with `UsePropertyNamingPolicy = true` exercising `FieldName()` under a naming policy. That edit is not part of this changeset.
- 305 unit + 249 integration + 49 Unit.Swagger + 17 Unit.Testing + 131 Int.Agents + 4 Int.OData passing. The Release build is clean (0 warnings) as of the final formatting pass; the suite runs above were taken immediately before it, so a re-run is worth doing before merge.

### Pre-existing failures, confirmed unrelated

Verified by stashing all changes and re-running:

- `Int.Swagger` 20/20 and `Int.OpenApi`.
- `EventQueueTests.event_hub_wakes_immediately_after_poll_timeout_when_new_event_arrives` is a timing flake — 2 of 4 baseline full-suite runs fail it, and it passes in isolation.

## Files

| File | Change |
|---|---|
| `Src/Library/Main/EndpointWarmup.cs` | new `PrecompileComplexBindGraph` warmup pass |
| `Src/Library/Binder/BinderExtensions.cs` | exclude indexers from `BindableProps`; cache `ElementIsCollection` |
| `Src/Library/Binder/ComplexSourceBinder.cs` | reject nested collections with a descriptive error |
| `Src/Attributes/Contracts/ReflectionCache.cs` | add `ElementIsCollection` |
| `Src/Generator/ReflectionGenerator.cs` | skip indexers when emitting property metadata |
| `Src/Library/Types.cs` | add `FromQueryAttribute` |
| `Src/Attributes/JetBrains/Annotations.cs` | `InternalsVisibleTo("Unit.FastEndpoints")` |
| `Tests/UnitTests/FastEndpoints/ComplexSourceBinderTests.cs` | new |
| `Tests/UnitTests/FastEndpoints/WarmupTests.cs` | +5 tests, graph-completeness sweep |


dj-nitehawk:
thank you very much 🙏

PR #1152, opened 2026-08-05
Updated the changelog mentioning the warmup changes

SjorsGoering:
#1151 added some improvements for the warmup logic and fixed an edge case. In that PR the changelog was not updated.
Let me know your thoughts regarding these changelog changes.


dj-nitehawk:
thank you!

PR #1154, opened 2026-08-09
Allocation-free EventBus dispatch + PublishFilteredAsync test coverage

SjorsGoering:
## Summary

`EventBus<TEvent>` dispatched every publish through `Parallel.ForEachAsync`, which is the most
expensive available way to run N (almost always 1) already-async handlers. This replaces it with
direct invocation, making the default publish path allocation-free.

`Parallel.ForEachAsync` allocated, per publish: internal `ForEachAsyncState`, a linked
`CancellationTokenSource`, a completion `Task`, worker tasks scheduled onto the pool, a display class
capturing the event model, and a redundant second async state machine per handler (the body lambda
was `async (h, c) => await h.HandleAsync(...)`, a pure passthrough).

## Changes

- `_handlers` is declared `IEventHandler<TEvent>[]` instead of `IEnumerable<>`. The backing store was
  already an array (see the existing `//ToArray() is essential here!!!`), so this only stops throwing
  the concrete type away.
- `!handlers.Any()` becomes `handlers.Length == 0` (no enumerator).
- `Mode.WaitForAll` invokes handlers directly: a fast path returning the single handler's task
  as-is, otherwise an indexed `Task[]` + `Task.WhenAll`.
- `Mode.WaitForAny` builds an indexed `Task[]` instead of a LINQ `Select`, and keeps `Task.WhenAny`.
- `Mode.WaitForNone` keeps offloading each handler via `Task.Run`.
- The ctor materializes a caller-supplied `IEnumerable` once instead of storing it. Previously the
  `.Any()` probe consumed an element of a non-replayable sequence and the collection was
  re-enumerated on every publish.

The `Task.Run` lambdas live in their own `Offload`/`OffloadAll`/`Filter` methods rather than inline.
Inlining makes Roslyn hoist the `<>c__DisplayClass` to the top of `Execute`, allocating 40 bytes on
**every** publish including the single-handler fast path. This is the same trap already recorded in
`.okf/gotchas.md` for `ValidationFailed`; a note has been added there for this call site too.

## Behavior changes

**Handlers are no longer throttled to `Environment.ProcessorCount`.** All handlers subscribed to an
event now start immediately on publish. Only observable for events with more subscribed handlers than
the machine has cores.

Everything else is preserved deliberately:

- `WaitForNone` stays fire and forget. Handlers are still offloaded, so a slow or *synchronously
  throwing* handler can neither block nor throw at the publisher. `IEventHandler.HandleAsync` is a
  plain `Task` method, and synchronous throws are real in practice, so inlining this branch would
  have been a semantic change against the documented `Mode.WaitForNone` contract.
- `WaitForAny` still returns a `Task.WhenAny(...)` rather than the single inner task, so it keeps
  completing rather than faulting when a handler throws.
- No public API changes.

## Measurements

`GC.GetTotalAllocatedBytes(precise: true)` over 500k publishes of an event whose handlers complete
synchronously, workstation GC, non-concurrent. An empty control loop measured 0 bytes/iteration.

| publish | before | after |
| --- | --- | --- |
| `WaitForAll`, 1 handler (the default) | 466.6 B | **0 B** |
| `WaitForAll`, 3 handlers | 598.4 B | 120 B |
| `WaitForNone`, 1 handler | 460.7 B | 272 B |
| `WaitForAny`, 1 handler | 568.0 B | 424 B |

## Tests

`PublishFilteredAsync` and the multi-handler `WaitForAll` branch had no coverage at all before this
change, despite both being rewritten here. Added to `Tests/UnitTests/FastEndpoints/EventBusTests.cs`:

- `WaitForAllInvokesEveryHandler` covers the new `Task.WhenAll` branch.
- `PublishFilteredOnlyInvokesMatchingHandlers` and `PublishFilteredWithNoMatchingHandlersIsNoOp`
  cover the filter path and its empty result.
- `WaitForNoneDoesNotSurfaceHandlerExceptions` pins the fire-and-forget contract: a
  synchronously-throwing handler must neither reach the publisher nor stop the remaining handlers.


dj-nitehawk:
thank you very much!

PR #1155, opened 2026-08-09
Reduce per-request allocations in SendBytesAsync and SendEmptyJsonObject

SjorsGoering:
## Reduce per-request allocations in `SendBytesAsync` and `SendEmptyJsonObject`

Two mechanical, behavior-neutral allocation fixes in `HttpResponseExtensions`. No public API changes.

### `SendBytesAsync`

It was a textbook redundant async wrapper: it `await`ed a single call that already returns
`Task<Void>`, so both the `async` state machine and the extra `Task<Void>` were pure overhead.
It now returns `SendStreamAsync`'s task directly.

The `using var memoryStream` also went away. `SendStreamAsync` already owns disposal via
`await using (stream)`, so the stream was being disposed twice; it is now disposed exactly once,
by the same code path that disposes every other stream response.

### `SendEmptyJsonObject`

It allocated a fresh `new JsonObject()` per call just to serialize `{}`. That is now a shared
`static readonly JsonObject`, matching the existing `_emptyObject` / `_emptyArray` precedent in
`Endpoint.Properties.cs`.

The response still goes through `SerOpts.ResponseSerializer` rather than writing a cached `"{}"`
literal, so a user's custom serializer hook still runs.

Note on sharing the instance: `JsonObject` lazily initializes its backing dictionary, so concurrent
first use can race, but both racers produce an empty dictionary and the serialized output is `{}`
either way.

### Measurements

500k iterations, `GC.GetTotalAllocatedBytes(precise: true)`, modelling the two shapes:

| | before | after |
|---|---|---|
| `SendBytesAsync` | 208 bytes/op | 136 bytes/op |
| `SendEmptyJsonObject` | 64 bytes/op | 0 bytes/op |


dj-nitehawk:
good one!
thanks!

PR #1158, opened 2026-08-13
Forwarded attribute decorations to every route of a multi-route endpoint

SjorsGoering:
# Forward attribute decorations to every route of a multi-route endpoint

## The bug

In `EndpointRouteMapper.Map()` the two per-definition cleanup lines sat *inside* the `foreach (var route in def.Routes)` loop instead of after it:

```csharp
                    totalEndpointCount++;
                }
                def.AttribsToForward = null;   // runs once per route
                def.IsLocked = true;           // runs once per route
            }
```

The indentation hides it. Both lines are already dedented relative to the verb loop, which makes them look like they escaped the route loop too, but they only escaped one level.

Two consequences:

1. **Forwarded attributes were lost on routes 2..N.** `AttribsToForward` collects every class-level attribute that FastEndpoints does not interpret itself (`EndpointExtensions.Initialize()`'s `default:` case), and the mapper copies it into the mapped endpoint's metadata so ASP.NET can see it. After route 1's iteration the field was null, so the `if (def.AttribsToForward is not null)` check failed for every remaining route.

   This is not just an OpenAPI cosmetic issue. Forwarded metadata is what drives framework middleware: `[EnableCors]`, `[EnableRateLimiting]`, `[RequestFormLimits]`, `[DisableRequestSizeLimit]`, `[RequireAntiforgeryToken]`, `[ApiExplorerSettings]`, plus any user or third party metadata attribute. In the repro below, `/one` gets the rate limiter and `/two` silently does not:

   ```csharp
   [HttpGet("/one", "/two")]
   [EnableRateLimiting("fixed")]
   sealed class MyEndpoint : EndpointWithoutRequest { ... }
   ```

2. **`IsLocked` was set a route early**, so the definition was read-only while routes 2..N were still being mapped. Latent rather than user visible today: every call in the per-route body only reads the definition and writes to the `RouteHandlerBuilder`, so nothing hits `ThrowIfLocked()`. It is a trap for anyone who later adds a definition-mutating call there, since it would fail only for multi-route endpoints.

Reachability of (1) is narrow, which is why it survived: `AttribsToForward` is only populated on the attribute-configuration path (the `else if` in `Initialize()`), so it needs attribute-based configuration **and** multiple routes on the `HttpAttribute`. No endpoint in the test harness used that combination, so nothing covered it.

The final state was correct either way: `Routes.Length == 0` throws during mapping, so the loop always runs at least once and every definition does end up locked. This is about ordering within mapping.

## The fix

Both lines dedented by one level, with a comment so the indentation trap does not get reintroduced.

## Tests

New `Tests/UnitTests/FastEndpoints/EndpointRouteMapperTests.cs`, following the existing `WarmupTests` pattern (isolated `WebApplication` plus `AddFastEndpoints([types])` so only the test endpoints are discovered, its own non-parallel collection, `Config.EpOpts.Filter` / `RoutePrefix` saved and restored):

- `ForwardedAttributes_AreAddedToEveryRouteOfAMultiRouteEndpoint`: an attribute-configured endpoint with two routes plus an unrecognized marker attribute. Asserts both mapped `RouteEndpoint`s carry the marker exactly once.
- `Definition_IsNotLockedUntilAllRoutesAreMapped`: a `Configure()`-based endpoint with two routes whose `Options(...)` lambda records `Definition.IsLocked` on each invocation. Asserts two recordings, both false, and that the definition is locked after mapping completes. Also asserts that all routes share one definition instance, which is why the lock state is checked once rather than per route.

Both were confirmed to fail against the unfixed mapper (the first on the second route's missing attribute, the second on the lock state) and to pass with the fix applied.

## Verification

- `Unit.FastEndpoints`: 313/313 (311 plus the 2 new).
- `Int.FastEndpoints`: 249/249.
- `dotnet build Src/Library/FastEndpoints.csproj -c Release`: clean for net8/net9/net10, 0 warnings.
- Full gate `dotnet test FastEndpoints.slnx -c Release --filter "ExcludeInCiCd!=Yes"`: green, `HttpFileExporterTests.plus_json_content_type_still_builds_json_skeleton`).
## OKF impact

- `Src/Library/changelog.md`: entry under Fixes, including the repro and a note that `Configure()`-based endpoints were never affected.
- `.okf/gotchas.md`: bullet recording that the per-definition cleanup must stay outside the route loop, why the indentation misleads, and where the coverage lives.
- No public API changes. `AttribsToForward` and `IsLocked` are both internal fields on `EndpointDefinition`, and neither is in the `InternalsVisibleTo` friend-assembly stock, so there is no `MissingMethodException` risk for `FastEndpoints.Mcp` / `FastEndpoints.A2A`.


dj-nitehawk:
nice catch... thanks!

PR #1159, opened 2026-08-13
Build the response cache 'Cache-Control' value once per endpoint

SjorsGoering:
# Build the response cache 'Cache-Control' value once per endpoint

## What

`ResponseCacheExecutor.Execute()` rebuilt an immutable string on every request to a response cached endpoint:

```csharp
cacheControlValue = $"{cacheControlValue}max-age={attrib.Duration}";
```

Both inputs are fixed after startup (`ResponseCache(...)` is `ThrowIfLocked()`-guarded, so `Location` and `Duration` cannot change once mapping is done), so this was a string interpolation plus an `int.ToString()` per request for a constant result.

## How

`ResponseCacheAttribute` is an MVC type we cannot extend, so the cached value lives on our own definition:

- New `internal string? CachedCacheControl` on `EndpointDefinition`, next to `ResponseCacheSettings`.
- `Execute` now takes the `EndpointDefinition` instead of the attribute and reads the settings off it. `FeRequestHandler` passes `epDef`.
- The location switch moved into a `BuildCacheControlValue(attrib)` helper, invoked through `epDef.CachedCacheControl ??= ...`.
- Only the `else` branch needed it. The `NoStore` branch already assigned constants and is untouched.

A race between the first concurrent requests to an endpoint is harmless: both compute the identical value and the reference assignment is atomic.

The emitted headers are unchanged, including the `Pragma: no-cache` side effect for `ResponseCacheLocation.None` and the bare `max-age=N` fallback for an out-of-range location value.

## Tests

This path had no coverage of the emitted header value at all, so new `Tests/UnitTests/FastEndpoints/ResponseCacheExecutorTests.cs` (12 tests) drives `Execute` directly against a `DefaultHttpContext` plus a faked `IResponseCachingFeature`:

- `Cache-Control` for `Any`, `Client`, `None`, and an out-of-range cast, each with its expected `Pragma`.
- Both `NoStore` combinations, including the `no-store,no-cache` append plus `Pragma` for `Location.None`.
- Vary-by header and vary-by query keys.
- Replacement of stale `Cache-Control` / `Pragma` / `Vary` left by earlier middleware.
- Endpoint without response caching configured (returns before the feature check), missing-middleware throw, zero-duration-without-`NoStore` throw.
- `CacheControlValue_IsBuiltOnceAndReusedAcrossRequests`: the second request receives the same string instance, which is the actual optimization.

Behavior parity was verified rather than assumed: with the old per-request interpolation temporarily restored, 11 of the 12 pass and only the build-once test fails. So the header assertions pin pre-existing behavior, and the caching assertion is a real regression test.

## Verification

- `Unit.FastEndpoints`: 323/323 (311 plus the 12 new).
- Full gate `dotnet test FastEndpoints.slnx -c Release --filter "ExcludeInCiCd!=Yes"`: `Unit.Testing` 17/17, `Int.Agents` 131/131, `Int.OData` 4/4, `Int.FastEndpoints` 244/244, `Unit.FastEndpoints` 323/323 and `HttpFileExporterTests.plus_json_content_type_still_builds_json_skeleton`).
- `dotnet build Src/Library/FastEndpoints.csproj -c Release`: clean for net8/net9/net10, 0 warnings.

## OKF impact

- `Src/Library/changelog.md`: entry under Improvements, noting the emitted headers are unchanged.
- `.okf/gotchas.md`: bullet recording that `CachedCacheControl` is only sound while response cache settings stay startup-fixed, so any future runtime mutation path (per-request duration, hot-reloadable settings, user code mutating the `ResponseCacheAttribute` reachable through endpoint metadata) must invalidate it.
- No public API changes. `CachedCacheControl` is a new internal field rather than a change to an existing member, and `ResponseCacheExecutor` is internal with `FeRequestHandler` as its only caller anywhere in `Src/`, so neither is in the `InternalsVisibleTo` notice surface for `FastEndpoints.Mcp` / `FastEndpoints.A2A`.


dj-nitehawk:
thank you very much!

Issue #1161, opened 2026-08-15
AddVersioning hides the IApiVersioningBuilder, blocking Asp.Versioning.OpenApi registration


SonnyRR:
## Description

The `AddVersioning` extension method of the `FastEndpoints.AspVersioning` package returns `IServiceCollection` instead of `IApiVersioningBuilder`. Internally it invokes `AddApiVersioning(...)` and `AddApiExplorer(...)`, which both return an `IApiVersioningBuilder`, but that builder is discarded because the method returns `services` directly.

This prevents consumers from chaining `AddOpenApi(...)` from the `Asp.Versioning.OpenApi` package, which registers the versioned `OpenAPI` document services, including the per-version document service and its `KeyedServiceContainer`. Without those registrations, the `OpenAPI` document endpoint mapped with `MapOpenApi(...)` and `WithDocumentPerVersion()` fails to resolve the per-version document service at request time.

## Configuration

The API is set up for URL-based versioning with `Asp.Versioning`, `Asp.Versioning.OpenApi`, `FastEndpoints` and `FastEndpoints.AspVersioning`:

```csharp
using Asp.Versioning;
using Asp.Versioning.ApiExplorer;
using Asp.Versioning.OpenApi;

using FastEndpoints;
using FastEndpoints.AspVersioning;

static void ConfigureApiVersioning(ApiVersioningOptions options)
{
    options.ReportApiVersions = true;
    options.AssumeDefaultVersionWhenUnspecified = true;
    options.ApiVersionReader = ApiVersionReader.Combine(new UrlSegmentApiVersionReader());
}

static void ConfigureApiExplorer(ApiExplorerOptions options)
{
    options.GroupNameFormat = "'v'VVV";
    options.SubstituteApiVersionInUrl = true;
}

void ConfigureOpenApi(VersionedOpenApiOptions options)
{
    options.Document.AddDocumentTransformer(
        (document, _, _) =>
        {
            document.Info.Title = apiOptions.Title;
            document.Info.Description = apiOptions.Description;

            return Task.CompletedTask;
        });
}

// The following no-op calls exist only to recover the discarded IApiVersioningBuilder,
// so that AddOpenApi(...) can be chained. This is a fragile workaround.
services
    .AddVersioning(ConfigureApiVersioning, ConfigureApiExplorer)
    .AddApiVersioning(_ => { })          // This method is implicitly invoked above.
    .AddApiExplorer(_ => { })            // This method is implicitly invoked above.
[...]

PR #1162, opened 2026-08-15
Add versioned OpenAPI options to AddVersioning

SonnyRR:
Extends `AddVersioning` with an optional `openApiOptions` parameter (available on `net10.0`) to configure versioned `OpenAPI` documents via `builder.AddOpenApi(...)`. Includes:
- Conditional `net10.0`-only parameter and `Asp.Versioning.OpenApi` package reference
- Unit tests for versioned `OpenAPI` configuration
- Changelog entry and `OKF` sync

Addresses #1161


dj-nitehawk:
merged...
thank you very much!

PR #1165, opened 2026-08-19
Faster idempotency policy metadata lookup

SjorsGoering:
## Summary

- Replaces `Metadata.OfType<EndpointDefinition>().SingleOrDefault()` with `Metadata.GetMetadata<EndpointDefinition>()` in `IdempotencyPolicy.CacheRequestAsync`, matching the lookup pattern already used in `AntiforgeryMiddleware`, `FeRequestHandler`, and `X402Middleware`.
- `IdempotencyPolicy` is registered as an ASP.NET output-cache base policy, so this lookup runs on every request in the app, not just idempotent endpoints. `OfType().SingleOrDefault()` allocates an enumerator and walks the whole metadata collection; `GetMetadata<T>()` is a cached, per-type lookup on `EndpointMetadataCollection`.
- Behavior-neutral: route mapping adds exactly one `EndpointDefinition` to metadata per endpoint, so `SingleOrDefault`'s throws-on-many semantics were never relied upon.


dj-nitehawk:
thank you!

PR #1169, opened 2026-08-24
don't assign null members when cloning a schema

### LINKED ISSUE #1168: OpenApi 8.3.0: schema clone assigns every null property, so nullable collections serialize `"const": null` and reject their own arrays
edit: Sorry about all the archeology below in the comments tracking down the root cause and whether or not this is intended behavior, etc.

## Bug

Nullable collection properties are emitted with a `const: null` sibling in 8.3.0:

```json
"children": {
  "const": null,
  "items": { "$ref": "#/components/schemas/Child" },
  "type": ["null", "array"]
}
```

There is no `oneOf`/`anyOf`/`allOf` in the node. JSON Schema applies keywords conjunctively, so the
property validates only when the value is `null` — a populated array fails against the document
describing it. Expected shape, per #1133:

```json
"children": { "type": ["null", "array"], "items": { "$ref": "#/components/schemas/Child" } }
```

## Root cause

`OpenApiSchema.Const` (`string?`) in `Microsoft.OpenApi` 2.11.0 tracks **assignment, not value** —
assigning `null` marks it present. Standalone, no FastEndpoints involved:

```csharp
var a = new OpenApiSchema { Type = JsonSchemaType.Array };            // {"type":"array"}
var b = new OpenApiSchema { Type = JsonSchemaType.Array };
b.Const = null;                                                       // {"const":null,"type":"array"}
```

`CloneConcreteSchemaFallback` (`Src/OpenApi/Schemas/OpenApiSchemaExtensions.cs:170`) assigns every
readable/writable property including nulls:

```csharp
property.SetValue(clone, CloneSchemaMemberValue(property.Name, property.GetValue(schema)));
```

so every fallback-cloned schema gets `Const` explicitly assigned null. Reproduced by replicating just
that loop against a hand-built array schema:

```
source           : {"type":"array","items":{"type":"string"}}
reflection clone : {"const":null,"type":["null","array"],"items":{"type":"string"}}
hand-built       : {"type":["null","array"],"items":{"type":"string"}}
```

Skipping null source values in that loop should be enough.

## Secondary: the fast path is dead on Microsoft.OpenApi 2.11.0

`OpenApiSchemaExtensions.cs:11` looks up a copy constructor that does not exist on that version:

Edit: The bug seems to originate in Microsoft.OpenApi 2.11.0, see 

```csharp
static readonly ConstructorInfo? _openApiSchemaCopyCtor = typeof(OpenApiSchema).GetConstructor([typeof(IOpenApiSchema)]);
```

`typeof(OpenApiSchema).GetConstructor([typeof(IOpenApiSchema)])` returns `null`, so
`CloneConcreteSchema` always takes the reflection fallback. That is presumably why the artifact
appears now: 8.3.0's nullable-array collapse routes these schemas through the clone for the first
time.

## Why 8.3.0 rather than 8.2.0

Holding FastEndpoints at 8.2.0 with `Microsoft.AspNetCore.OpenApi` 10.0.11 yields zero occurrences;
moving only FastEndpoints to 8.3.0 yields one per nullable collection property. The two releases fail
in opposite directions — 8.2.0's `oneOf` form has no null branch so a `null` fails, while 8.3.0
rejects non-null arrays — and with `DefaultIgnoreCondition = WhenWritingNull` the property is omitted
rather than written as `null`, so 8.3.0 is the more disruptive of the two.

## Repro

```csharp
public sealed record Child(int Value);
public sealed record Parent(IReadOnlyList<Child>? Children);

sealed class GetParent : EndpointWithoutRequest<Parent>
{
    {
        Get("/parent");
        AllowAnonymous();
    }

    public override Task HandleAsync(CancellationToken ct)
        => Send.OkAsync(new Parent([new Child(1)]), ct);
}
```

`GET /openapi/v1.json` → `components.schemas.Parent.properties.children` carries `const: null`.

## Environment

- `FastEndpoints.OpenApi` 8.3.0
- `Microsoft.AspNetCore.OpenApi` 10.0.11
- `Microsoft.OpenApi` 2.11.0
- .NET 10

StepKie:
Fixes #1168.

`OpenApiSchema.Const` tracks *assignment* rather than value since `Microsoft.OpenApi` 2.11.0 —
microsoft/OpenAPI.NET#2935 asked for that deliberately, since `const: null` is valid under JSON Schema
2020-12. `CloneConcreteSchemaFallback` assigned every readable/writable member including nulls, so an
untouched `Const` became an explicit one on the clone and the writer emitted `"const": null`.

The visible symptom is nullable collection properties, which 8.3.0 newly routes through this clone:

```json
"children": { "const": null, "items": { "$ref": "…/Child" }, "type": ["null","array"] }
```

Keywords apply conjunctively, so that only validates when the value *is* null — a populated array
fails against its own schema.

Skipping null source values is behaviour preserving: the clone starts as a fresh `OpenApiSchema` with
every member unset, so not assigning a null leaves it exactly where it already was.

Two things to point out:

- The test passes with or without the fix while this repo resolves `Microsoft.OpenApi` 2.7.5
  transitively through `Microsoft.AspNetCore.OpenApi` 10.0.11, because assigning null is harmless
  there. It starts failing at 2.11.0 or later.
- `_openApiSchemaCopyCtor` is always null, so this fallback is the only clone path in practice:
  `typeof(OpenApiSchema).GetConstructor([typeof(IOpenApiSchema)])` returns null on both 2.7.5 and
  2.11.0. Left alone here, since which constructor is intended is your call.


dj-nitehawk:
thank you!
`v8.4.0-beta.2` has been pushed to nuget with the fix.

PR #1171, opened 2026-09-01
Fix URL encoding of route values in routeless test helpers

Benziza:
## Summary

URL-encode hydrated route parameter values generated by `GetTestUrlFor()`.

Route values containing reserved URI characters such as `#`, `?`, or spaces are currently inserted directly into the generated URL. This can truncate or alter the request before it reaches the endpoint.

## Reproduction

Given the following route:

```csharp
Get("/items/{name}");
```

And this request value:

```csharp
Name = "part#1?x y";
```

`GetTestUrlFor()` currently generates:

```text
items/part#1?x y
```

The actual request path becomes:

```text
/items/part
```

## Fix

Encode hydrated route parameter values with `Uri.EscapeDataString()` while preserving unresolved route placeholders.

Catch-all route handling is intentionally outside the scope of this change.

## Tests

- [x] `HttpClientExtensionsTests`: 12/12 passed
- [x] `Unit.FastEndpoints`: 346/346 passed
- [x] Full solution build: 0 warnings and 0 errors


Benziza:
Companion documentation PR: https://github.com/FastEndpoints/Documentation/pull/40

dj-nitehawk:
thank you!

PR #1174, opened 2026-09-09
Add RabbitMQ command and event transport

axies20:
## Summary

- add a new `FastEndpoints.Messaging.RabbitMQ` package for one-way `ICommand` and `IEvent` delivery
- add typed consumer registration using the standard FastEndpoints command and event handler contracts
- add publisher confirms with a bounded pool of exclusive, long-lived channels via `PublisherConcurrency`
- add manual ACK/NACK handling, scoped handler resolution, configurable concurrency and prefetch
- support Classic, Quorum, and Stream queue declarations with custom topology arguments
- support producer-only route registration and Aspire-style named connection strings

## API

```csharp
builder.AddRabbitMQMessaging(
           "messaging",
           options => options.PublisherConcurrency = 8)
       .AddCommand<CreateOrder, CreateOrderHandler>()
       .AddEvent<OrderCreated, OrderCreatedHandler>();
```

Producer-only applications can register routes without starting local consumers:

```csharp
builder.AddRabbitMQMessaging("messaging")
       .AddCommandPublisher<CreateOrder>()
       .AddEventPublisher<OrderCreated>();
```

## Behavior

- each active publication exclusively leases one channel until its publisher confirm completes
- the pool creates at most `PublisherConcurrency` channels and reuses them across publications
- waiting for a channel is asynchronous and cancellable
- failed publications return their pool slot, and shutdown waits for active leases
- ordering is preserved with the default concurrency of `1`; ordering across multiple channels is not guaranteed
- `ICommand<TResult>` request/reply is intentionally outside the scope of this transport

## Validation

- `16` RabbitMQ unit tests passed
- `6` broker integration cases passed against RabbitMQ 4, including `32` concurrent publications through a pool of `4` channels
- package build passed for `net8.0`, `net9.0`, and `net10.0` with `PublicSign=true`, with `0` warnings and `0` errors
- `git diff --check` passed

The broader existing messaging regression suite could not be rebuilt locally because RSA strong-name signing fails under the machine's OpenSSL policy with `OpenSslCryptographicException: invalid digest`. The new package and its focused test projects build successfully with public signing.

Public release notes and the repository OKF knowledge set are included. The sibling `FE-Docs` repository was not available in the local workspace.


dj-nitehawk:
Thanks for the PR but this is not something I'd want to be maintaining. You can host the package yourself as a community addon and we can link to it from the CookBook.

It does need more work before it's ready for production though...

## Correctness Bugs

**1. Successful work is requeued on shutdown**

`RabbitMQConsumerService` acks with `lifetime.ApplicationStopping`. If the handler finishes and shutdown has already started, `BasicAckAsync` throws `OperationCanceledException`, and this catch treats completed work as incomplete:

```csharp
catch (OperationCanceledException) when (lifetime.ApplicationStopping.IsCancellationRequested)
{
    await AcknowledgeAsync(() => SafeNackAsync(channel, args.DeliveryTag, requeue: true));
}
```

That is duplicate delivery on every graceful stop. Ack completed work with `CancellationToken.None` (or any token that is not `ApplicationStopping`). Only nack + requeue when the handler itself was cancelled.

**2. Producer-only events fail when nothing is bound**

`PublishCore` always uses `mandatory: true`. Event publisher routes set `Queue = null`, so the publisher never binds a queue. RabbitMQ.Client 7.2.2 throws `PublishException` on `basic.return`.

Commands are fine (the publisher declares the queue). `AddEventPublisher<T>()` is in the PR body, changelog, and unit tests, and it will throw at runtime if no consumer is currently bound. Use `mandatory: false` for events, or only mark commands mandatory.

**3. Event publish topology is mixed with per-handler queue settings**

`AddEvent` stores `route with { Queue = null, AutoDelete = false }`, which still carries consumer `QueueArguments` (`x-queue-type`, DLX, …). `PublisherTopologyMatches` compares those fields.

- `.AddEventPublisher<T>().AddEvent<T, H>()` throws: empty args vs `{ x-queue-type: classic }`
- Two handlers of the same event with different `QueueType` / `QueueArguments` throw, even though each handler has its own queue

Publisher matching should only cover exchange, type, routing key, durability, and exchange args. Store `RabbitTopology.Event<TEvent>(...)` as the publish route.

## Also needed

- Consumer `catch` swallows handler/JSON failures with no `ILogger` and no hook, and `RequeueOnFailure` defaults to `false`. Poison messages disappear. Remote messaging logs and has an exception receiver. At least log.
- `StopAsync` disposes channels without `BasicCancel` and without waiting for in-flight `ReceivedAsync`.
- `AutomaticRecoveryEnabled` plus `RabbitConnection.GetAsync` disposing/recreating when `!IsOpen` fight each other during recovery.
- Missing tests: producer-only event publish with no consumers; `AddEventPublisher` + `AddEvent`; two handlers with different queue args; graceful shutdown after a successful handler.

PR #1175, opened 2026-09-13
Make the test url cache route opt-in

mbcrawfo:
## Problem

`EndpointRouteMapper.Map()` unconditionally maps `GET /_test_url_cache_`. The route is anonymous, excluded from OpenAPI, and returns every registered endpoint as `{EndpointType.FullName}|{route template}`. Endpoint-level auth doesn't apply to it, so in a production app any caller can list all routes, including authorized endpoints and endpoints hidden from Swagger/OpenAPI docs. The only mitigation today is a global `AuthorizationOptions.FallbackPolicy`.

The route only exists so routeless test helpers (`GETAsync<TEndpoint>()` etc.) can resolve URLs when the app under test runs out of process (Aspire, Native AOT).

## Change

- The route is mapped only when the configuration value `FastEndpoints:ExposeTestUrlCache` is `true` (e.g. env var `FastEndpoints__ExposeTestUrlCache=true`).
- `GetTestUrlFor()` now throws an `InvalidOperationException` explaining how to enable it when the route returns 404, instead of a bare `HttpRequestException`.
- The Native AOT `AppFixture` sets `FastEndpoints__ExposeTestUrlCache=true` by default, so AOT tests need no changes.
- In-process WAF tests are unaffected (they read the in-memory cache directly).
- Changelog entry under "Minor Breaking Changes": Aspire tests must set the value on the tested project resource:

```csharp
appHost.CreateResourceBuilder<ProjectResource>("apiservice")
       .WithEnvironment("FastEndpoints__ExposeTestUrlCache", "true");
```

A matching docs update is ready for the Documentation repo.

## Testing

- New integration test: `/_test_url_cache_` returns 404 on the default test app.
- New unit test: the out-of-process fallback throws the actionable error on 404.
- `dotnet test --solution FastEndpoints.slnx -c Release`: 992/994 passed. The 2 failures are `Int.OpenApi.Kiota` failing to start the Debug harness executable in a Release run; that project passes 4/4 when run on its own.
- `NativeAotCheckerTests` in AOT mode: 48/48 passed (exercises the fixture default end to end).


dj-nitehawk:
thank you very much 🙏

PR #1176, opened 2026-09-13
Match the Bearer scheme case-insensitively in JwtRevocationMiddleware

mbcrawfo:
## Summary

`JwtRevocationMiddleware` only runs `JwtTokenIsValidAsync()` when the `Authorization` header starts with `Bearer ` using a case-sensitive (and culture-sensitive) `StartsWith`. `JwtBearerHandler` matches the scheme with `StringComparison.OrdinalIgnoreCase`, and RFC 7235 defines auth schemes as case-insensitive.

As a result, a revoked but still unexpired token sent as `Authorization: bearer <jwt>` (or `BEARER`, `bEaReR`, ...) skips the revocation check entirely, yet is still accepted by the authentication handler and reaches the endpoint as an authenticated user.

## Reproduction

With `app.UseJwtRevocation<T>()` configured and a blacklisted token:

| Header | Before | After |
|---|---|---|
| `Bearer <jwt>` | 401 `Bearer token has been revoked!` | 401 `Bearer token has been revoked!` |
| `bearer <jwt>` | 200, `JwtTokenIsValidAsync` never called | 401 `Bearer token has been revoked!` |
| `BEARER <jwt>` | 200, `JwtTokenIsValidAsync` never called | 401 `Bearer token has been revoked!` |

## Changes

- Match the prefix with `StringComparison.OrdinalIgnoreCase`, consistent with `JwtBearerHandler`.
- `SecurityTests.Jwt_Revocation` is now a theory over `Bearer`, `bearer` and `BEARER`. The latter two fail without the fix.
- Changelog entry under Fixes.


dj-nitehawk:
thank you!

PR #1177, opened 2026-09-13
Always bind the permission check result to optional [HasPermission] props

mbcrawfo:
## Summary

Optional `[HasPermission]` properties could be set by the client through the JSON request body.

`RequestBinder` deserializes the JSON body into the whole request DTO before the other binding sources run. For `[HasPermission(isRequired: false)]` properties, `BindHasPermissionProps` only wrote the value when the user held the permission. When the permission was missing, nothing overwrote the property, so a value sent in the body survived.

## Reproduction

```csharp
public class UpdateItemRequest
{
    public int Id { get; set; }
    public decimal NewPrice { get; set; }

    [HasPermission("Inventory_Update_Price", isRequired: false)]
    public bool CanUpdatePrice { get; set; }
}
```

An authenticated user **without** the `Inventory_Update_Price` permission sends:

```http
POST /update
Content-Type: application/json

{ "id": 1, "newPrice": 1, "canUpdatePrice": true }
```

`req.CanUpdatePrice` is `true` in the handler. With the permission present, a body value of `false` is overwritten to `true`, so the binder was only authoritative in one direction.

## Fix

`BindHasPermissionProps` now writes the permission check result unconditionally, so the property is `false` whenever the permission is missing, regardless of the request body. The `isRequired: true` validation failure path is unchanged.

Behavior change: optional `bool?` permission properties are now `false` instead of `null` when the permission is missing.

## Tests

New `RequestBinderTests` cases:

- body `true`, permission missing: binds `false`
- body `false`, permission present: binds `true`
- optional `bool?`, permission missing: binds `false`

`Unit.FastEndpoints` and `Int.FastEndpoints` pass locally.

## Docs

Changelog entry added. Companion docs PR: FastEndpoints/Documentation#42


dj-nitehawk:
Thanks for the writeup and tests. I am going to reject this as a fix, because the behavior you are closing is the designed meaning of `isRequired: false`.

## What `isRequired: false` is for

`IsRequired` is a validation switch, not a “this property is claims-only” switch.

The attribute docs only talk about the 400:

> set to false if a validation error shouldn't be thrown when the current principal doesn't have the specified permission.

The class summary only documents the `true` write:

> boolean properties decorated with this attribute will have their values set to **true** if the current principal has the specified permission.

They never say “write `false` when the permission is missing.” Current model-binding docs match that: disable the automatic validation error, nothing about forcing `false` or ignoring the body.

Binding order is JSON → form → route → query → claims → headers → permissions. Later sources overwrite earlier ones **only when they write**. The old `BindHasPermissionProps` is the same pattern as optional `[FromClaim]` / `[FromHeader]` / `[FromCookie]`:

| Permission | `isRequired` | Result |
| --- | --- | --- |
| present | either | `true` (body `false` loses) |
| missing | `true` | validation failure |
| missing | `false` | leave whatever JSON / query / default already set |

So: if the permission is present, take that value regardless of JSON. If it is not present and validation is off, take the JSON (or query, or the CLR default). That is the matrix this PR is deleting.

## This is not an accident in the rest of the stack

Swagger/OpenAPI treat optional HasPermission as a client-supplied field on purpose.

Required permission props are stripped from the request body. Optional ones stay in the schema unless `RemoveFromSchema` is set. The NSwag processor is explicit:

```csharp
case HasPermissionAttribute pAttrib:
    return !pAttrib.IsRequired; // add param if it's not required. if required only can bind from actual permission.
```

Test helpers follow the same rule: required HasPermission is skipped when building JSON / route values; optional is not.

If optional HasPermission was never supposed to come from the client, it would already be removed from the body schema the way `isRequired: true` is. `RemoveFromSchema` is the existing knob for “keep it off the contract.”

## The `CanUpdatePrice` sample is the wrong tool

```csharp
[HasPermission("Inventory_Update_Price", isRequired: false)]
public bool CanUpdatePrice { get; set; }
```

then trusting `req.CanUpdatePrice` as authz in the handler is a misuse of an optional bind. For “does this user have the permission?” use `User.HasPermission(...)`, or `[HasPermission]` with the default `isRequired: true`. Do not put an authorization flag on the DTO and then ask the binder to make it unforgeable **only when the claim is missing**.

On `bool`, the CLR default is already `false` when the client omits the property. The only extra behavior `isRequired: false` adds is: the client *may* send a value when they do not have the permission. That is the feature.

## What this PR would break

- Anyone using optional HasPermission as permission-or-body fallback (the documented binding-order model).
- Optional `bool?` properties: missing permission goes from `null` to `false`.
- OpenAPI / Swagger would still advertise the property on the body (and as a query param on bodyless verbs) while the binder silently ignored it.

The new gotcha that trusted-source binders “must write unconditionally” is the same overreach. Apply that to optional `[FromClaim]` and you delete that fallback too.

## If you want claims-only without a 400

That is a new behavior, not a bugfix. It would need to be opt-in (or a documented breaking change), and OpenAPI / Swagger / test helpers would have to stop treating `isRequired: false` as a client-bindable field.

I am not taking that change in this form. Closing this PR. The companion docs PR should be closed with it.

PR #1178, opened 2026-09-13
Check the accepted token and unrouted requests in JwtRevocationMiddleware

mbcrawfo:
## Summary

`JwtRevocationMiddleware` has two cases where `JwtTokenIsValidAsync()` is never called for a token that still authenticates:

1. **Tokens not sent in the `Authorization` header.** The middleware only reads the header, but apps commonly also accept the access token from the query string or a cookie via `JwtBearerEvents.OnMessageReceived` (the standard pattern for SSE, EventSource and SignalR clients). A revoked token sent as `?access_token=<jwt>` is accepted by `JwtBearerHandler` and reaches the endpoint without a revocation check. If `OnMessageReceived` lets the query token take precedence, a client can even send a different, non-revoked token in the header to pass the header check while the handler authenticates the revoked one.
2. **No matched endpoint.** `GetEndpoint()?.Metadata.OfType<IAllowAnonymous>().Any() is null or true` treats a null endpoint as allow-anonymous. When `UseJwtRevocation<T>()` runs before an explicit `UseRouting()` call, every request skips the check, with no error or log.

## Reproduction

`[Authorize]` endpoint, `app.UseJwtRevocation<T>()`, `OnMessageReceived` reading `?access_token=`, and a blacklisted but still valid token:

| Setup | Request | Before | After |
|---|---|---|---|
| Correct middleware order | `Authorization: Bearer <jwt>` | 401 revoked | 401 revoked |
| Correct middleware order | `?access_token=<jwt>` | 200, check never called | 401 revoked |
| `UseJwtRevocation` before `UseRouting` | `Authorization: Bearer <jwt>` | 200, check never called | 401 revoked |
| `UseJwtRevocation` before `UseRouting` | `?access_token=<jwt>` | 200, check never called | 401 revoked |

## Changes

- Only endpoints with `IAllowAnonymous` metadata skip the check. Requests with no matched endpoint are checked.
- After the existing header check, the middleware calls `AuthenticateAsync` for each scheme whose handler is a `JwtBearerHandler` and checks the `access_token` saved on a successful result (if it differs from the header token). Authentication results are cached per request, so the token is not validated twice and the documented order (before `UseAuthentication`) still works.
- `AddAuthenticationJwtBearer` now defaults `JwtBearerOptions.SaveToken` to `true` (set before the user callback, so it can still be turned off). Apps that call `AddJwtBearer()` directly need `SaveToken = true` for non-header tokens to be checked; otherwise behavior is unchanged from today.
- No public API changes. `JwtTokenIsValidAsync` and `SendTokenRevokedResponseAsync` are unchanged, and the header check still runs for tokens the handler rejects.
- Tests: new `JwtRevocationMiddlewareTests` unit tests (null endpoint, anonymous skip, query token, valid header plus revoked query token, single validation per request) and integration test `SecurityTests.Jwt_Revocation_Token_From_Query_String`. The harness now accepts `?access_token=`. The bypass tests fail without the fix.
- Changelog entry under Fixes and an OKF gotcha.

Docs: FastEndpoints/Documentation#44


dj-nitehawk:
thank you very much!

PR #1179, opened 2026-09-13
Build ServiceResolver singletons from the root provider instead of the request scope

mbcrawfo:
## Summary

`ServiceResolver.CreateSingleton` builds the instances FastEndpoints caches as singletons (validators, mappers, event handlers, command handler executors) from `HttpContext.RequestServices` when they are first created during a request. That is the default, since they are created lazily unless `Warmup()` is enabled.

The instance is cached for the process lifetime, but its constructor dependencies come from the scope of whichever request created it first. A scoped dependency such as a `DbContext` or a current user service is captured from that first request and reused by every later request. DI scope validation never fires, because the scoped service really is resolved from a request scope and the caching happens outside the container. The docs already say that injecting a scoped service into a validator constructor throws, but that is only true with `Warmup()`.

## Reproduction

A scoped `ICurrentUser` reads its name from an `X-User` header. An event handler and a validator both constructor-inject it. Two requests are sent, `X-User: alice` then `X-User: bob`, without `Warmup()`:

| | Before | After |
|---|---|---|
| `Resolve<ICurrentUser>()` in the endpoint (control) | alice, bob | alice, bob |
| Development (scope validation on) | handler and validator see alice, alice, no error | `InvalidOperationException: Cannot resolve scoped service 'ICurrentUser' from root provider.` |
| Production (scope validation off) | handler and validator see alice, alice | alice, alice (same captive dependency behavior as plain ASP.NET Core) |

## Changes

- `CreateSingleton` builds from the root provider of the host serving the request, which is what `Warmup()` already does. That root is taken from the host's own `IServiceResolver` singleton in `RequestServices`, not from the captured `provider`: in multi-host test runs the static `ServiceResolver.Instance` can belong to a disposed host (the same problem #1118 fixed for `TryResolve`). Outside a request the captured root is used as before.
- Unit test mode (`Factory.AddTestServices`) keeps using the http context services, since its root provider is empty.
- `ServiceResolverTests`: replaced `CreateSingleton_UsesHttpContextRequestServicesIfAvailable` with tests showing that a scoped ctor dependency throws under scope validation inside a request (captured root and live host root variants), that the live host root is used when the captured root is disposed, and that unit test mode still uses request services. The two scoped dependency tests fail without the fix.
- Changelog entry under Fixes, `IServiceResolver.CreateSingleton` XML doc, and an OKF gotcha.

Unit.FastEndpoints, Unit.Testing, Int.FastEndpoints, Int.OpenApi and Int.Agents all pass locally.

Docs: FastEndpoints/Documentation#45


dj-nitehawk:
thank you!

